PR #827, opened 2016-11-18
Use MarionetteDriver for Firefox 48+

Artur-:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/827)

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/BrowserUtil.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/BrowserUtil.java
@@ -243,12 +243,14 @@ public class BrowserUtil {
             } else if (p == Platform.MAC) {
                 return "Mac";
             }
-            
+
         } catch (Exception e) {
         }
-        Object rawPlatform = capabilities.getCapability(CapabilityType.PLATFORM);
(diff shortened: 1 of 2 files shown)


alvarezguille:
can you add a test even if it's only possible to run locally?

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java, line 56 at r1](https://reviewable.io:443/reviews/vaadin/testbench/827#-KWrQD0Dyz-Um8PEJlQV:-KWrQD0Dyz-Um8PEJlQW:bf5hql8) ([raw file](https://github.com/vaadin/testbench/blob/cb17066f61f43f596ad8413bcda6f268a831ccc0/vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java#L56)):*
> ············
?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/827#-:-KWrRCsWf_O8DHBHLDLX:b-ezp6q4)*

Artur-:
Seems like this does not work reliably with Selenium 2 and GeckoDriver says you really should use Selenium 3, so let's upgrade to 3 instead

PR #832, opened 2016-11-21
Make it possible to selectByText in ComboBoxes with text input disabled

### LINKED ISSUE #758: ComboBoxElement selectByText does not work correctly
**Originally by _rhaex_**

___
When using a ComboBoxElement to control the behaviour of a ComboBox during integration tests, the value selecting does not work correctly when the ComboBox has setTextInputAllowed set to false.

It uses javascript to remove the read only flag, type in the selection and then re enable the read only flag. This changes the value in the ComboBox but does not trigger a value change event, and when you continue the value returns to the original value.

___
 Imported from https://dev.vaadin.com/ issue #15316

Artur-:
From the 4.2 branch

Fixes #758

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/832)


alvarezguille:
Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/832)*

PR #853, opened 2016-12-09
Click on buttons using space in IE

Artur-:
This should resolve most problems with IE misclicking and not
hitting buttons at all

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/853)

### DIFF
--- a/doc/releasenotes.html
+++ b/doc/releasenotes.html
@@ -492,6 +492,10 @@ img {
     <li>waitForVaadin handles a possible null return value</li>
     <li>The hub host name can be overridden using -Dcom.vaadin.testbench.Parameters.hubHostname</li>
     <li>The browser name and version to run on locally can be overridden using -Dcom.vaadin.testbench.Parameters.runLocally</li>
+    <li>Buttons in Internet Explorer are primarily clicked by pressing space
+    on them to work around problems in the Internet Explorer web driver. The
+    feature can be disabled using
+    -Dcom.vaadin.testbench.Parameters.ieClickWorkaround=false</li>
     
     <li>
         Component Element API changes in vaadin-testbench-api 7.7.4:
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/Parameters.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/Parameters.java
@@ -12,6 +12,8 @@
  */
 package com.vaadin.testbench;
 
+import org.openqa.selenium.WebElement;
+
 import com.vaadin.testbench.annotations.RunLocally;
 import com.vaadin.testbench.annotations.RunOnHub;
 
@@ -25,6 +27,7 @@ public class Parameters {
     private static int screenshotRetryDelay = 500;
     private static int testsInParallel;
     private static int testSuitesInParallel;
+    private static boolean ieClickWorkaround;
     static {
         isDebug = getSystemPropertyBoolean("debug", false);
 
@@ -49,6 +52,8 @@ public class Parameters {
         } else {
             testsInParallel = 50;
         }
+
+        ieClickWorkaround = getSystemPropertyBoolean("ieClickWorkaround", true);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


rogozinds:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java, line 195 at r1](https://reviewable.io:443/reviews/vaadin/testbench/853#-KYmBicGJFzZYlH8QP0m:-KYmBicGJFzZYlH8QP0n:b-qjg931) ([raw file](https://github.com/vaadin/testbench/blob/19e5493bea239b6305f8640e83422f80be5546cc/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java#L195)):*
> ```Java
>         if (BrowserUtil.isIE(getCapabilities())) {
>             // IE randomly misclicks because of problems in the IE driver.
>             // This is a workaround for Vaadin buttons and native buttons, which
> ```

Why it's not in the ButtonElement class? I guess for support "native" buttons, but that's inconsistent, because other methods doesn't support native buttons.

Another issue is that ButtonElement has a hack for buttons with style hidden and this is a hack for IE, will be nice just to have them in one place.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/853)*

Artur-:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java, line 195 at r1](https://reviewable.io:443/reviews/vaadin/testbench/853#-KYmBicGJFzZYlH8QP0m:-KYmcX9CLOFy7jpLgJep:b-waadz) ([raw file](https://github.com/vaadin/testbench/blob/19e5493bea239b6305f8640e83422f80be5546cc/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java#L195)):*
<details><summary><i>Previously, rogozinds (rogozinds) wrote…</i></summary><blockquote>

Why it's not in the ButtonElement class? I guess for support "native" buttons, but that's inconsistent, because other methods doesn't support native buttons.

Another issue is that ButtonElement has a hack for buttons with style hidden and this is a hack for IE, will be nice just to have them in one place.
</blockquote></details>

All WebElements are TestBenchElements unless you specifically wrap them as something else. It would be weird if `click` worked differently depending on how you retrieve the element reference.

The ButtonElement hack should probably then be in this method also, if it should work also if you do something like
`findElement(By.id("mybutton")).click()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/853)*

Artur-:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java, line 199 at r1](https://reviewable.io:443/reviews/vaadin/testbench/853#-KYmct-mnIlRvKPXajAc:-KYmct-mnIlRvKPXajAd:b-fh6p58) ([raw file](https://github.com/vaadin/testbench/blob/19e5493bea239b6305f8640e83422f80be5546cc/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java#L199)):*
> ```Java
>             if (hasClassName("v-button")
>                     || "button".equalsIgnoreCase(getTagName())) {
>                 actualElement.sendKeys(Keys.SPACE);
> ```

For some cases, this could potentially cause problems (e.g. somebody has globally mapped space to a short cut or some other weird thing). There should thus be a way to skip this behavior

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/853)*

Artur-:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java, line 199 at r1](https://reviewable.io:443/reviews/vaadin/testbench/853#-KYmct-mnIlRvKPXajAc:-KYn6UVAFLBWWaB3USkZ:bwtp71v) ([raw file](https://github.com/vaadin/testbench/blob/19e5493bea239b6305f8640e83422f80be5546cc/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchElement.java#L199)):*
<details><summary><i>Previously, Artur- (Artur) wrote…</i></summary><blockquote>

For some cases, this could potentially cause problems (e.g. somebody has globally mapped space to a short cut or some other weird thing). There should thus be a way to skip this behavior
</blockquote></details>

Added a system property for this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/853)*

alvarezguille:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/853)*

Artur-:
Let's skip this for now as the Framework project indicates that if you are running with `requireWindowFocus`, then `sendKeys` will not work properly

PR #856, opened 2016-12-16
Define the core version again in the Framework API module

Artur-:
This ensures that the Framework API classes are compatible with
a release version and not only a snapshot

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/856)


Artur-:
No longer needed as the module was moved to the framework project

PR #884, opened 2017-03-31
Add gson dependency to avoid problems with JBrowserDriver

### LINKED ISSUE #872: java.lang.NoClassDefFoundError: com/google/gson/JsonSyntaxException when adding dependency on jbrowserdriver
TestBench uses `gson` directly for license checking. However, the `gson` dependency is not a direct dependency of TestBench, but instead an indirect dependency (of selenium-api 2.53.1). This dependency breaks and is removed when a dependency on `jbrowserdriver` is introduced (since it excludes gson from its dependency on selenium-api to 2.53.0 for unknown reason).

Easily reproducible:

1. git clone https://github.com/johannesh2/testbench4-starterkit
2. Add dependency to `pom.xml`:
```xml
    <dependency>
      <groupId>com.machinepublishers</groupId>
      <artifactId>jbrowserdriver</artifactId>
      <version>0.17.3</version>
      <scope>test</scope>
    </dependency>
```
3. Run `MyFirstIT` class. There is no need to launch a Jetty server to reproduce this issue, TestBench will fail sooner.

Artur-:
Fixes #872

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/884)

### DIFF
--- a/vaadin-testbench-core/pom.xml
+++ b/vaadin-testbench-core/pom.xml
@@ -221,5 +221,10 @@
             <artifactId>gwt-elemental</artifactId>
             <version>2.8.0</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.3.1</version>
+        </dependency>
     </dependencies>
 </project>


alvarezguille:
@Artur- is this still needed after #880 ?

Artur-:
Seems like it is not needed for 5.0-SNAPSHOT

PR #911, opened 2017-06-08
Automatically scroll elements into view before interacting with them

### LINKED ISSUE #895: Automatically scroll elements into view before interacting with them
The webdriver spec says
```
14. Element Interaction

The element interaction commands provide a high-level instruction set for manipulating form controls. Unlike Actions, they will implicitly scroll elements into view and check that it is an interactable element.
```

Seems like this is not how Selenium implements e.g. `click`, as the [javadoc](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/WebElement.java#L43) says `There are some preconditions for an element to be clicked. The element must be visible and it must have a height and width greater then 0.`

Right now if you try to click on an element outside the viewport, you get
```
org.openqa.selenium.ElementNotVisibleException: element not visible
```

To deal with this, TestBench could automatically trigger `scrollIntoView` for the element before you do actions such as `click` or `getText` on it - similarly to how `waitForVaadin` works.

alvarezguille:
Fixes #895

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/911)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,8 @@ vaadin-testbench-recorder/chrome/testbench-recorder.jar
 vaadin-testbench-recorder/extension/
 vaadin-testbench-recorder/install.rdf
 vaadin-testbench-playground
-testbench-integration-tests/error-screenshots/
-testbench-integration-tests/config/testbench.properties
+vaadin-testbench-integration-tests/error-screenshots/
+vaadin-testbench-integration-tests/config/testbench.properties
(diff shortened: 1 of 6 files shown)


Artur-:
Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/911)*

PR #965, opened 2017-11-21
add unique Sauce Connect tunnel ID for each build

### LINKED ISSUE #959: Define a tunnel identifier for sauce connect
Without this, running one build with cancel terminate the tunnel for another build. It could probably be set in pom.xml for the Sauce connect plugin using e.g. `maven.build.timestamp` and passed on to the integration tests and then set as a `DesiredCapability`

vlukashov:
Pass a unique value (the maven build timestamp) as a tunnel id to the sauce connect plugin to ensure two builds started from two machines around the same time would not interfere with one another.

+ switch the flow dependency from SNAPSHOT to a specific version

fixes #959

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/965)

### DIFF
--- a/vaadin-testbench-integration-tests/pom.xml
+++ b/vaadin-testbench-integration-tests/pom.xml
@@ -23,6 +23,8 @@
         <node.version>v8.1.2</node.version>
         <yarn.version>v0.27.5</yarn.version>
         <buildtools.directory>build-tools</buildtools.directory>
+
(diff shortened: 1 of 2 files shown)


alvarezguille:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/vaadin/testbench/965)*

PR #1071, opened 2018-05-17
 Return all elements found by selector in all Vaadin clients

### LINKED ISSUE #756: TestBench query fails when multiple vaadin portlets are on a single page
**Originally by _roflkoptr_**

___
I have 2 vaadin portlets on a single page in liferay portal. Each has its own UI and a Button component in it.
When testing the page, I am trying to find both buttons on the page using this query:
```
$(ButtonElement.class).all();
```
I expect the query to return both buttons in a list. But only the button from the first portlet is returned.
On the other hand, querying each of the buttons separately using findElement(By.vaadin(...)) works fine. So I made the following simple test to see what works and what doesn't:
```
@Test
public void findAllButtons() {          //fails
	assertTrue($(ButtonElement.class).all().size() == 2);
}

@Test
public void findFirstButton() {         //passes
	$(ButtonElement.class).id("button_1").click();
}

@Test
public void findSecondButton(){         //fails
	$(ButtonElement.class).id("button_2").click();
}

@Test
public void findElementByVaadin(){       //passes
	findElement(By.vaadin("//com.vaadin.ui.Button[caption='button_2']"));
}

@Test
public void findElementsByVaadin(){     //fails
	findElements(By.vaadin("//com.vaadin.ui.Button[caption='button_2']")).get(0);
}
```

Source of the portlets and test is in attachment.

My speculation (might be totally wrong): 
From the results of the test, it seems that findElement works fine but findElements returns only results from the first portlet on the page. After quickly browsing through the code I think that I found the problem.
In method TestBenchDriverProxy#executeSearch, the script fragment created by:
```_ Construct elementSelectionString script fragment based on type of
_ search context
String elementSelectionString # "var elementclients[client]."
        + jsFunction;
if (context instanceof WebDriver) {
	elementSelectionString += "(arguments[0]);";
} else {
	elementSelectionString += "StartingAt(arguments[0], arguments[1]);";
}

String findByVaadinScript # "var clientswindow.vaadin.clients;"
		+ "for (client in clients) {" + elementSelectionString
		+ "  if (element) {" + " return element;" + "  }" + "}"
		+ "return null;";
```
does not work for jsFunctions getElementsByPath and getElementsByPathStartingAt. They return an empty array when nothing is found, so the "if (element) {" part of the script is always true. getElementByPath and getElementByPathStartingAt return null when nothing is found, so it works for them.

___
 Imported from https://dev.vaadin.com/ issue #15295

DiegoCardoso:
Change JS logic to accumulate all elements found by Vaadin clients instead of returning only for the first client

Fixes #756

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/1071)

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchDriverProxy.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBenchDriverProxy.java
@@ -118,7 +118,7 @@ public class TestBenchDriverProxy extends TestBenchCommandExecutor implements
 
     /**
      * Finds a list of elements by a Vaadin selector string.
-     * 
+     *
      * @param selector
      *            TestBench4 style Vaadin selector.
      * @param context
@@ -148,9 +148,10 @@ public class TestBenchDriverProxy extends TestBenchCommandExecutor implements
         }
 
         String findByVaadinScript = "var clients = window.vaadin.clients;"
+                + "var elements = [];"
                 + "for (client in clients) {" + elementSelectionString
-                + "  if (element) {" + " return element;" + "  }" + "}"
-                + "return null;";
+                + "  if (element) {" + " elements = elements.concat(element);" + "  }" + "}"
+                + "return elements;";
 
         WebDriver driver = ((HasDriver) context).getDriver();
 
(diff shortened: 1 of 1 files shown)


DiegoCardoso:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/vaadin/testbench/1071)*

PR #1157, opened 2019-04-04
Update code style to Vaadin convention

oluwasayo:
- Format Java and XML files to Vaadin convention
- Remove dead blocks
- Optimize POM files
- Collapse redundant intermediate modules
- Rename modules

---
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/1157)


alvarezguille:
we should remove the dependency to vaadin-bom
and rename the folder with the artifact id, at least for the root level folders

PR #1170, opened 2019-04-15
Check license headers

oluwasayo:
This change is [Reviewable](https://reviewable.io/reviews/vaadin/testbench/1170)


oluwasayo:
I've now skipped the check on those files as they are test sources which we don't ship.

PR #1266, opened 2020-06-10
feat: add HasHelper interface

DiegoCardoso:
Interface to be implemented by field components which has the helper feature.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
     </parent>
     <artifactId>vaadin-testbench-parent</artifactId>
     <packaging>pom</packaging>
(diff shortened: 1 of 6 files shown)


DiegoCardoso:
There's a [currently draft PR](https://github.com/vaadin/vaadin-radio-button-flow/pull/140) for Radio Button where this is being tested at the first scenario.

For the later, I tested locally in Select, which has a slotted slot to check if the actual element we want was being returned and could confirm it.

DiegoCardoso:
Done. Now there's a `6.3` branch.

PR #1276, opened 2020-11-03
Updated license to latest Vaadin Commercial Developer License

### LINKED ISSUE #1283: Update license to CVDL4
First, bump the version to 7.0-SNAPSHOT.
Similarly to MPR, we need to change the license for TB. This can be done with a plugin, see https://github.com/vaadin/multiplatform-runtime-internal/commit/ece90326ab40b5b198ef04908904262df6baff4c
(make sure it is not triggered as part of the build, but can be done manually).

mstahv:
Fixes #1283

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,169 +1,262 @@
 Commercial Vaadin Developer License version 4
+
 Terms and Conditions for Use, Reproduction and Distribution
 
-NOTICE TO USER: PLEASE READ THIS LICENSE AGREEMENT CAREFULLY. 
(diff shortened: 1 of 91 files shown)


alvarezguille (dismissed):
`You should have received a copy of the license along with this program.` was removed in other jar artifacts.
Also remember to bump to next major version.

mstahv:
I wonder why was it removed, we don't package the license in the jar files?

I'm still also wondering if we should change the inception year in the source headers 🤔

alvarezguille:
I think that it was never packaged in the jar, it was included in the zip that users downloaded from directory

caalador (approved):

PR #1287, opened 2021-03-18
fix: support jdk 16 by aligning bytebuddy version in flow

manolo:
Related with https://github.com/vaadin/flow/issues/10115

### DIFF
--- a/vaadin-testbench-core/pom.xml
+++ b/vaadin-testbench-core/pom.xml
@@ -212,11 +212,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <!-- Override transitive dependency with Java 11 compatible version -->
+        <!-- Override transitive dependency with Java 16 compatible version -->
         <dependency>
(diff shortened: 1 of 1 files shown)


vaadin-testbench-core/pom.xml
```diff
@@ -216,7 +216,7 @@
         <dependency>
             <groupId>net.bytebuddy</groupId>
             <artifactId>byte-buddy</artifactId>
-            <version>1.9.13</version>
+            <version>1.10.21/version>
```
Artur-, line 219: Looks good if you add the missing `<`

vaadin-testbench-core/pom.xml
```diff
@@ -216,7 +216,7 @@
         <dependency>
```
Artur-, line 216: Also update the comment above to say Java 16

Artur- (changes requested):

Artur- (approved):

PR #1298, opened 2021-09-09
Override outdated guava (transient dep of remote driver)

fluorumlabs:
This updates guava to more recent version


mshabarov:
@fluorumlabs do you remember what is the transitive version of guava currently used? I want to check the difference between them, before merging.

fluorumlabs:
The one brought by Selenium is 25.0-jre

mshabarov (approved):

PR #1313, opened 2022-02-04
Use slf4j like the rest of the platform

Artur-:


alvarezguille:
@Artur- should `testbench-core` now have an explicit dependency to slf4j?

Artur-:
yes

alvarezguille (approved):

PR #1318, opened 2022-02-10
chore: Move shared functionality out from testbench-core

Artur-:
Shared is defined as functionality that depends on Selenium but not on the testing framework you use


Artur-:
rebased

PR #1343, opened 2022-05-09
chore: Fix deprecation warnings

alvarezguille:
Use Browser instead of BrowserType
Use getPlatformName instead of getPlatform
Use getBrowserVersion instead of getVersion

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/BrowserUtil.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/BrowserUtil.java
@@ -11,7 +11,6 @@ package com.vaadin.testbench.parallel;
 
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.Platform;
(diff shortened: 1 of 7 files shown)


vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/BrowserUtil.java
```diff
@@ -87,7 +86,7 @@ public static boolean isIE(Capabilities capabilities) {
         if (capabilities == null) {
             return false;
         }
-        return BrowserType.IE.equals(capabilities.getBrowserName());
+        return Browser.IE11.name().equals(capabilities.getBrowserName());
```
Artur-, line 89: Is there any point in retaining IE11 stuff when we do not support IE anymore?

Artur-: https://github.com/vaadin/testbench/issues/1347

PR #1345, opened 2022-05-11
chore: Added support for Kotlin sources

### LINKED ISSUE #1328: Setup kotlin compilation for Core module
To support bringing in the kotlin JUnit features into the project kotlin compilation using the `kotlin-maven-plugin` should be setup for the `vaadin-testbench-core` module

mcollovati:
## Description

Configured kotlin-maven-plugin to allow compilation of Kotlin source code along with Java sources.
Replaced maven-javadoc-plugin with dokka-maven-plugin in core module to generate Javadocs also for Kotlin classes.
Attached kotlin source files to sources jar.

Fixes #1328 

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.

### DIFF
--- a/vaadin-testbench-core/pom.xml
+++ b/vaadin-testbench-core/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


caalador (approved):
Seems ok.
Formatting line length for xml files could be longer so we could have elements on 
single line like, but that's just a preference for reading the xml.

vaadin-testbench-core/pom.xml
```diff
@@ -87,12 +117,106 @@
         </resources>
 
         <plugins>
+            <plugin>
```
caalador, line 120: There could actually be a note here that the kotlin plugin needs to be before the maven-compiler-plugin so that the kotlin files are compiled before the java compiler runs so that no-one moves the definitions around as a mistake.

Another would be to set the phase for kotlin so that it always executes before the java compileer.

mcollovati: I would prefer adding a comment that explains the need to have kotlin plugin before maven-compiler-plugin, rather than using a phase different from `compile` for the kotlin plugin

caalador (approved):

PR #1346, opened 2022-05-11
feat: Added base classes for JUnit UI unit tests 

### LINKED ISSUE #1330: Create abstract UITest class
Create a base class for using UI JUnit tests.
This should handle all the basic setup and teardown of test mocks and Route registration.

Acceptance criteria
- Can create test by extending interface and UI.getCurrent() is available.
- Documentation for Base setup is available

mcollovati:
## Description

Added base classes for UI unit test for both JUnit 4 and JUnit 5.
Current implementation depends on Karibu testing framework.

Fixes #1330 

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] I have added a description following the guideline.
- [X] The issue is created in the corresponding repository and I have referenced it.
- [X] I have added tests to ensure my change is effective and works as intended.
- [X] New and existing tests are passing locally with my change.
- [X] I have performed self-review and corrected misspellings.

- [X] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


vaadin-testbench-core/src/main/java/com/vaadin/karibu/BaseKaribuTest.java
```diff
@@ -0,0 +1,59 @@
+/**
+ * Copyright (C) 2020 Vaadin Ltd
```
caalador, line 2:
```suggestion
 * Copyright (C) 2022 Vaadin Ltd
```
Should the year be updated to current year at least for the new headers?

vaadin-testbench-core/src/main/java/com/vaadin/karibu/BaseKaribuTest.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * This program is available under Commercial Vaadin Developer License
+ * 4.0 (CVDLv4).
+ *
+ *
+ * For the full License, see <https://vaadin.com/license/cvdl-4.0>.
+ */
+package com.vaadin.karibu;
```
caalador, line 10:
```suggestion
package com.vaadin.testbench.unit;
```
I think we should not have karibu in the package name as it should just be another way of tetsting with TestBench. 
Perhaps use `unit`, `junit` or `ui` to keep some clarity between class use cases.

vaadin-testbench-core/src/main/java/com/vaadin/karibu/Karibu5Test.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * Base JUnit 5 class for UI unit tests.
+ * 
+ * Subclasses should typically restrict classpath scanning to a specific package
+ * for faster bootstrap by overriding {@link #scanPackage()} method.
+ */
+public abstract class Karibu5Test extends BaseKaribuTest {
```
caalador, line 21: Again in the naming of the abstract I would perhaps not use Karibu in the name but leave it as an implementation detail.

Also there is a target of going to junit5 as default so I would have this be the base name `UnitTest`? and have the junit4 version have the version in the name `Unit4Test`

caalador: Same for the test names and packages. to not use karibu.

caalador (changes requested):

vaadin-testbench-core/src/main/java/com/vaadin/karibu/Karibu5Test.java
```diff
@@ -0,0 +1,34 @@
...
+package com.vaadin.karibu;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+/**
+ * Base JUnit 5 class for UI unit tests.
+ * 
+ * Subclasses should typically restrict classpath scanning to a specific package
+ * for faster bootstrap by overriding {@link #scanPackage()} method.
```
caalador, line 19: Should we add a note here that if the extending class overrides the init/clean methods the '`@BeforeEach` annotation needs to be added there as the superclass annotation is not inherited in junit5 if the method is overridden.

vaadin-testbench-core/src/main/java/com/vaadin/karibu/BaseKaribuTest.java
mcollovati: `com.vaadin.testbench.unit` was my first choice, then I changed it but only because `com.vaadin.karibu` was used in sample code.
I'll revert the packages to `com.vaadin.testbench.unit`

vaadin-testbench-core/src/main/java/com/vaadin/karibu/Karibu5Test.java
mcollovati: What about `UIUnitTest`? It may better explain the purpose of the base class.

caalador (approved):

PR #1348, opened 2022-05-13
feat: add navigation methods to base

### LINKED ISSUE #1331: Add navigation to abstract test class
Implement navigate methods in the UITest class for navigating to a view.
The navigation should return the view instance navigated to `T navigate(Class<T> navigationTarget)`
e.g. navigate(MyView.class) should return the MyView instance attached to the UI.

For String and other navigation methods the return value will probably be of type HasElement.

Acceptance:
- methods for navigating using each available UI.navigate is implemented
- Navigating to a view is documented.

caalador:
Add methods for navigating to
different views into the base class.

Fixes #1331

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -73,7 +73,7 @@ class BaseUIUnitTest {
      */
     public <T extends Component> T navigate(Class<T> navigationTarget) {
         UI.getCurrent().navigate(navigationTarget);
-        return (T) getCurrentView();
+        return navigationTarget.cast(getCurrentView());
     }
 
     /**
@@ -92,7 +92,7 @@ class BaseUIUnitTest {
     public <C, T extends Component & HasUrlParameter<C>> T navigate(
             Class<T> navigationTarget, C parameter) {
         UI.getCurrent().navigate(navigationTarget, parameter);
-        return (T) getCurrentView();
+        return navigationTarget.cast(getCurrentView());
     }
 
     /**
@@ -110,7 +110,7 @@ class BaseUIUnitTest {
     public <T extends Component> T navigate(Class<T> navigationTarget,
             Map<String, String> parameters) {
         UI.getCurrent().navigate(navigationTarget, new RouteParameters(parameters));
-        return (T) getCurrentView();
+        return navigationTarget.cast(getCurrentView());
     }
 
     /**
@@ -135,7 +135,7 @@ class BaseUIUnitTest {
                             + currentView.getClass().getName() + " instead of "
                             + expectedTarget.getName());
         }
-        return (T) currentView;
+        return expectedTarget.cast(currentView);
     }
 
     /**
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -9,16 +9,21 @@
  */
 package com.vaadin.testbench.unit;
 
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import com.github.mvysny.kaributesting.v10.MockVaadin;
 import com.github.mvysny.kaributesting.v10.Routes;
 
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.component.HasElement;
 import com.vaadin.flow.component.UI;
+import com.vaadin.flow.router.HasUrlParameter;
+import com.vaadin.flow.router.RouteParameters;
 
 /**
  * Base class for UI unit tests.
- * 
+ *
  * Provides methods to set up and clean a mocked Vaadin environment.
  *
  * Subclasses should typically restrict classpath scanning to a specific package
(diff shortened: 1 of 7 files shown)


vaadin-testbench-core/src/test/java/com/example/base/About.java
```diff
@@ -17,4 +17,15 @@
 @Tag("div")
 @Route("about")
 public class About extends Component {
+    long id = System.currentTimeMillis();
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof About && this.id == ((About) obj).id;
+    }
```
mcollovati, line 25: Why is a custom `equals` implementation needed? Isn't identity check not enough?

vaadin-testbench-core/src/test/java/com/example/base/About.java
```diff
@@ -17,4 +17,15 @@
...
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
+    }
```
mcollovati, line 30: Can be removed, since it only invokes super? Or is the method overridden for some specific purpose?

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -56,4 +61,90 @@ protected void cleanVaadinEnvironment() {
...
+     *
+     * @param navigationTarget
+     *         view class to navigate to
+     * @param <T>
+     *         view type
+     * @return instantiated view
+     */
+    public <T extends Component> T navigate(Class<T> navigationTarget) {
+        UI.getCurrent().navigate(navigationTarget);
+        return (T) getCurrentView();
```
mcollovati, line 76: What about using `navigationTarget.cast(getCurrentView())`, so IntelliJ IDEA stops to complain about unchecked cast?
```suggestion
        return navigationTarget.cast(getCurrentView());
```

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -56,4 +61,90 @@ protected void cleanVaadinEnvironment() {
...
+     * @param parameters
+     *         parameters to pass to view.
+     * @param <T>
+     *         view type
+     * @return instantiated view
+     */
+    public <T extends Component> T navigate(Class<T> navigationTarget,
+            Map<String, String> parameters) {
+        UI.getCurrent().navigate(navigationTarget, new RouteParameters(parameters));
+        return (T) getCurrentView();
```
mcollovati, line 113: Same as above
```suggestion
        return navigationTarget.cast(getCurrentView());
```

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -56,4 +61,90 @@ protected void cleanVaadinEnvironment() {
...
+            Class<T> expectedTarget) {
+        UI.getCurrent().navigate(location);
+        final HasElement currentView = getCurrentView();
+        if (!expectedTarget.equals(currentView.getClass())) {
+            throw new IllegalArgumentException(
+                    "Navigation resulted in unexpected class "
+                            + currentView.getClass().getName() + " instead of "
+                            + expectedTarget.getName());
+        }
+        return (T) currentView;
```
mcollovati, line 138: Same as above
```suggestion
        return expectedTarget.cast(currentView);
```

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/UIUnit4BaseClassTest.java
```diff
@@ -57,13 +57,17 @@ public void extendingBaseClass_runTest_vaadinInstancesAvailable() {
...
+                    getCurrentView() instanceof Home);
         }
 
     }
 
     public static class DiscoverAllRoutesTest extends UIUnit4Test {
+        @Override
+        protected String scanPackage() {
+            return "com.example.base";
+        }
```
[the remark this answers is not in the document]
caalador: The test becomes an upkeep test if it is for all as we probably want to add new routes for different reasons. But I'll update them for this round.

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/UIUnitBaseClassTest.java
```diff
@@ -51,15 +51,19 @@ void extendingBaseClass_runTest_vaadinInstancesAvailable() {
...
         }
 
     }
 
     @Nested
     class DiscoverAllRoutesTest extends UIUnitTest {
+        @Override
+        protected String scanPackage() {
+            return "com.example.base";
+        }
```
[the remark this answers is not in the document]
caalador: updated.

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/UIUnit4BaseClassTest.java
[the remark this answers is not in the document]
mcollovati: The test is there because base class has **null** as default for `scanPackage` method, but it may not be a good default.
Perhaps we can change it in another PR with a better value, because it is unlikely that developer needs to scan full classpath; test class package may be a better choice.

PR #1353, opened 2022-05-16
chore: Imported Karibu kotlin sources

### LINKED ISSUE #1329: Import required kotlin classes
Bring in [karibu-tools](https://github.com/mvysny/karibu-tools), [mock-servlet-environment](https://github.com/mvysny/karibu-testing/tree/master/mock-servlet-environment), [karibu-testing-v10](https://github.com/mvysny/karibu-testing/tree/master/karibu-testing-v10) Kotlin code into core module.

add the `io.github.classgraph:classgraph` dependency.

For the imported files any vaadin version checks that do different things for 14, 15, 19+ should be removed during import as the target version is only 23+ and the plaform version should always be the supported version. Any required version checks should utilize FrontendVersion instead of VaadinVersion.kt as the semver parse can not handle for instance 23.1-SNAPSHOT

Initial import should target utils, mocks and flow-server features. 
Component extensions should be taken later as needed.

mcollovati:
## Description

Imported kotlin source code and test from karibu.
Part of the added code and tests may be dropped as Java APIs evolves.
Some source code and tests are commented and marked with as TODO,
because they are strictly related to specific component or to Servlet 4 APIs

Fixes #1329


vaadin-testbench-core/pom.xml
```diff
@@ -378,28 +378,53 @@
             <artifactId>kotlin-stdlib</artifactId>
             <version>${kotlin.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
```
caalador, line 383: kotlin-stdlib-jdk8 brings kotlin-stdlib with it so the kotlin-stdlib can be removed from dependencies

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/ComponentUtils.kt
```diff
@@ -0,0 +1,393 @@
...
+var Component.textAlign: String?
+    get() = element.style.get("textAlign")
+    set(value) { element.style.set("textAlign", value) }
+
+/**
+ * Sets or removes the `title` attribute on component's element.
+ */
+var Component.tooltip: String?
+    get() = element.getAttribute("title")
+    set(value) { element.setOrRemoveAttribute("title", value) }
```
caalador, line 85: This sounds a bit strange as an addition for testing as the feature doesn't exist.

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/BasicUtils.kt
```diff
@@ -0,0 +1,178 @@
...
+fun Component.isEffectivelyEnabled(): Boolean = isEnabled
+
+/**
+ * Checks whether this component is [HasEnabled.isEnabled]. All components not implementing [HasEnabled] are considered enabled.
+ */
+val Component.isEnabled: Boolean
+    get() = when (this) {
+        is HasEnabled -> isEnabled
+        else -> true
+    }
```
caalador, line 135: This is totally skiping the fact that the check should be `component.getElement().isEnabled()` which is basically what the mixin gets as a shorthand.
This should be removed as it makes things wrong as a non HasEnabled that is in a disabled component should return that it is disabled.

mcollovati:
Tests pass locally with Java 17

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/BasicUtils.kt
```diff
@@ -0,0 +1,178 @@
...
+        val hasValue = this as HasValue<HasValue.ValueChangeEvent<Any?>, Any?>
+        check(!hasValue.isReadOnly) { "The ${toPrettyString()} is read-only" }
+    }
+}
+
+/**
+ * Fails if the component is editable. See [checkEditableByUser] for more details.
+ * @throws AssertionError if the component is editable.
+ */
+fun Component.expectNotEditableByUser() {
```
caalador, line 108: We should drop any `expect...` methods, but you can create a ticket for that.

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/ComponentUtils.kt
caalador: perhaps drop any extension that adds get set for attributes.

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/ComponentUtils.kt
```diff
@@ -0,0 +1,393 @@
...
+    // workaround for https://github.com/vaadin/flow/issues/11709
+    style.clear()
+    addClassNames2(*classNames)
+}
+
+/**
+ * A component placeholder, usually shown when there's no value selected.
+ * Not all components support a placeholder; those that don't will return null.
+ */
+var Component.placeholder: String?
```
caalador, line 269: We should have these extensions that are not for all components so that the feature is findable for testing.

caalador: Also for the components that have a placeholder the method is setPlaceholder so this wouldn't really give anything new on that part.

caalador (approved):
Can just generate tickets about the comments so that we clean some of the unwanted component methods and component references away.

mcollovati:
> Can just generate tickets about the comments so that we clean some of the unwanted component methods and component references away.

I'll fix POM immediately and create tickets for the other comments

vaadin-testbench-core/pom.xml
mcollovati: done

caalador (approved):

vaadin-testbench-core/src/main/kotlin/com/vaadin/testbench/unit/internal/BasicUtils.kt
mcollovati: added ticket #1354

PR #1362, opened 2022-05-17
feat: Added notification wrapper

### LINKED ISSUE #1335: Create test wrapper for Notification
Extend the generic component wrapper for Notification.

Notification should contain at least the getText test method that validates the the component is seemingly usable and return the text shown in the notification.

Initially add the extending wrappers to TB instead of the component TestBench modules which is the targeted place.

Acceptance criteria:
- Wrapper exists with tests
- Documentation on getting/using wrapper

mcollovati:
Adds a wrapper for notification with methods to get notification text and
to close a displayed notification.

Fixes #1335

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -215,4 +215,16 @@ class BaseUIUnitTest {
                             .getSimpleName());
         }
     }
+
+    /**
+     * Simulates a server round-trip, flushing pending component changes.
+     */
+    protected static void roundTrip() {
+        UI.getCurrent().getInternals().getStateTree()
+                .collectChanges(nodeChange -> {
+                });
+        UI.getCurrent().getInternals().getStateTree()
+                .runExecutionsBeforeClientResponse();
+    }
+
 }
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
@@ -81,7 +81,7 @@ public class ComponentWrap<T extends Component> {
      */
     public void setModal(boolean modal) {
         UI.getCurrent().setChildComponentModal(component, modal);
-        flushChanges();
+        roundTrip();
     }
 
     /**
@@ -100,12 +100,8 @@ public class ComponentWrap<T extends Component> {
     /**
      * Simulates a server round-trip, flushing pending component changes.
      */
-    static void flushChanges() {
-        UI.getCurrent().getInternals().getStateTree()
-                .collectChanges(nodeChange -> {
-                });
-        UI.getCurrent().getInternals().getStateTree()
-                .runExecutionsBeforeClientResponse();
+    protected void roundTrip() {
+        BaseUIUnitTest.roundTrip();
     }
 
 }
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
@@ -13,6 +13,7 @@ import org.slf4j.LoggerFactory;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.flow.component.UI;
+import com.vaadin.testbench.unit.internal.PrettyPrintTreeKt;
 
 /**
  * Test wrapper for components with helpful methods for testing a component.
@@ -21,7 +22,7 @@ import com.vaadin.flow.component.UI;
  * wrappers.
  *
  * @param <T>
- *         component type
+ *            component type
  */
 public class ComponentWrap<T extends Component> {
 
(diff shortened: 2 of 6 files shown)


vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
```diff
@@ -21,7 +22,7 @@
  * wrappers.
  *
  * @param <T>
- *         component type
+ *            component type
```
caalador, line 25: Seems there is a difference in our formatters. Could fix this so the un-needed moving front and back doesn't happen.

mcollovati: I'm using Eclipse Code Formatter within IntelliJ IDEA with Flow repo rules (eclipse/VaadinJavaConventions.xml).
Perhaps there's still something wrong on my side.
I'll check and fix it, thanks.

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/NotificationWrapTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        Assertions.assertFalse(notification_.isUsable(),
+                "Not opened Notification shouldn't be usable");
+    }
+
+    @Test
+    void openedNotification_isUsable() {
+        String notificationText = "Opened notification";
+        Notification notification = Notification.show(notificationText);
+        ComponentWrap.flushChanges();
```
caalador, line 39: Would be nice if we could somehow run this automatically when ever something is set to `beforeClientResponse`, but don't know how complex that would be.

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
```diff
@@ -75,14 +76,36 @@ private boolean isEffectivelyVisible(Component component) {
...
+            throw new IllegalStateException(
+                    PrettyPrintTreeKt.toPrettyString(component)
+                            + " is not usable");
+        }
+    }
+
+    /**
+     * Simulates a server round-trip, flushing pending component changes.
+     */
+    static void flushChanges() {
```
caalador, line 103: This might need to be protected as the component wrappers when moved to components repository might end up under `com/vaadin/flow/component/[componentName]/testbench/` like the elements.

Also this could almost be named `roundTrip` and be added to the BaseUnitTest class so that users could also use this when a simulated roundtrip would be needed.

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/NotificationWrap.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * If the notification is not displayed an IllegalStateException will be
+     * thrown as the end user would not be able to see the contents.
+     * 
+     * @return the text of the notification
+     * @throws IllegalStateException
+     *             is notification is not displayed
+     */
+    public String getText() {
+        ensureComponentIsUsable();
```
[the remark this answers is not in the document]
caalador: Actually should the notification isUsable only be about isAttached and isOpened?

caalador (changes requested):

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/NotificationWrapTest.java
caalador: added note for this

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/NotificationWrap.java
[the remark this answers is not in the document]
mcollovati: It makes sense to consider only isAttached and isOpened.
User can interact with Notification only if it contains interactive elements, and the tests should do the same.
It the notification is for some reason disabled, also children components will be, and their wrapper will prevent interactions.
Furthermore, also `close` method in wrapper may be removed, since there is no way to close a notification other than waiting for auto-close or having some component or shortcut that invokes `Notification.close`

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
mcollovati: Renamed and added the method to BaseUnitTest. ComponentWrap also has a protected `roundTrip` method that delegates to BaseUnitTest. Hope I correctly understood your comment.

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/NotificationWrap.java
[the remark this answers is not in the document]
mcollovati: Would add also `isVisible` to the check

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -215,4 +215,16 @@ private <T extends ComponentWrap<Y>, Y extends Component> T initialize(
                             .getSimpleName());
         }
     }
+
+    /**
+     * Simulates a server round-trip, flushing pending component changes.
+     */
+    protected static void roundTrip() {
```
caalador, line 222: Let's keep this protected for now and try to automate where ever it is needed, but it might be that we need this as public api.

caalador (approved):

PR #1364, opened 2022-05-18
chore: added maven formatter plugin

mcollovati:


caalador (commented):
Run the formatter once for the sources also so we have the formatting.

caalador (approved):

PR #1365, opened 2022-05-18
chore: add formatter plugin

caalador:
Format sources with formatter.


caalador:
Closing in favor of #1364

PR #1367, opened 2022-05-18
feat: Automatically select matching wrapper

### LINKED ISSUE #1352: Extend getting test wrapper
When getting the test wrapper it should be possible to receive a component specific wrapper instead of the default `ComponentWrap`.

Test wrappers should get a `Targets`/`Wraps` (or other name) annotation that can contain an array of base classes that the wrapper specifically handles.

The annotation is then used to determine which wrapper the given component should use.

Wrapper classes should be under a specific package so that we can have a fast scan for all available warppers.

caalador:
Fixes #1352

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -9,12 +9,20 @@
  */
 package com.vaadin.testbench.unit;
 
+import javax.servlet.annotation.HandlesTypes;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -9,12 +9,20 @@
  */
 package com.vaadin.testbench.unit;
 
+import javax.servlet.annotation.HandlesTypes;
+import java.util.Arrays;
```
mcollovati, line 13: Unused imports
```suggestion
```

mcollovati:
apart from imports,  LGTM

mcollovati (approved):

Issue #1370, opened 2022-05-19
Add more component query helpers.


caalador:
Create methods in ComponentQuery for all possible LocatorSpecifications.

- [x] caption #1384 
- [x] ~~placeholder~~ (see #1357)
- [x] text #1385 
- [x] count: exact, range, max, min #1374 
- [x] value #1388 
- [x] applied CSS classes #1373 
- [x] not applied CSS classes #1373

knoobie:
Just putting the ideas here; not sure if you already wanna have separate tickets in this early stage. 

You often use the theme attribute to apply custom styling to components, so it would be awesome to search for them, like you allow with classes (that aren't so common with components) (mentioned in the other PR)

Additionally it would be good to add generic hasAttribute(String attribute) and hasAttribute(String attribute, String value) to search components by custom attributes like aria-label, aria-hidden and so on.

mcollovati:
Thanks again for your feedback.
Do you mind open two separate issues for theme and attribute helpers?

PR #1373, opened 2022-05-19
feat: Added component query helpers for CSS classes

mcollovati:
Added helper for searching for components with and without CSS classes
applied.

Part of #1370

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentQuery.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentQuery.java
@@ -10,8 +10,8 @@
 package com.vaadin.testbench.unit;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ComponentQuery.java
```diff
@@ -114,6 +118,56 @@ public ComponentQuery<T> withCondition(Predicate<T> condition) {
...
+     * Requires the components to have all the given CSS class names
+     *
+     * @param className
+     *            required CSS class name, not {@literal null}
+     * @param other
+     *            additional required CSS class names
+     *
+     * @return this element query instance for chaining
+     */
+    public ComponentQuery<T> withClassName(String className, String... other) {
```
knoobie, line 131: Should there be a note that the order did matter or is it irrelevant? Looking below with the Hashset and String.join that looks like a culprit for unexpected behavior

mcollovati: The order here doesn't matter, since under the hood every single class name is checked  separately

knoobie: Thanks! With that in mind - would it make sense to add a test that 

withClassName("test-class other-class") would work as well as little convenient for the developer? There were similar additions to flow in the past to easy development with multiple css classes

caalador (approved):

PR #1381, opened 2022-05-20
feat: Add query by theme

### LINKED ISSUE #1375: [uitest] Query components by theme attribute
As a developer using UI tests I want the ability to query components by the theme attribute.  

The theme attribute is used by components to have different variants and it's the recommended way to apply custom styling to components, so it would be awesome to search for them,already possible with classes that aren't so common with field.

caalador:
Added possibility to define query to
requre themes or tht they are not available.

Fixes #1375


mcollovati (commented):
Comments are mainly about cosmetics, so LGTM also as is

mcollovati (approved):

PR #1384, opened 2022-05-20
feat: added component query helper for caption

mcollovati:
Added helpers to search for components with caption matching or containing given test

Part of #1370


mcollovati:
The `caption` extension in ComponentUtils.kt is marked as deprecated and not to use.
To be checked if we shall use the actual implementation, trying something different or drop the concept of  `caption` in favor of `label`, that is effectively known by several components through `HasLabel` mixin

caalador (approved):

caalador:
Some conflicts arose due to remove public

PR #1385, opened 2022-05-20
feat: Added component query helper for text content

mcollovati:
Added helpers to search for components with text content equals or containing
given text.

Part of #1370


mcollovati:
For `contains` condition I added a Java predicate, because I had troubles invoking kotlin extension function `textContains` function from Java due to generics bounds.

vaadin-testbench-core/src/main/java/com/vaadin/testbench/unit/ElementConditions.java
```diff
@@ -0,0 +1,107 @@
...
+                componentText = component.getElement().getText();
+            } else {
+                return false;
+            }
+            if (componentText == null) {
+                return false;
+            }
+            // WARN: may not work correctly with unicode chars
+            if (ignoreCase) {
+                return componentText.toLowerCase().contains(text.toLowerCase());
```
knoobie, line 102: Should this check be done with Locale.ROOT on both sides to make sure the developer machine's default locale won't interfere with the results? I've seen reports in the past with turkish letters that were problematic

mcollovati: Good catch, thanks!

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/ComponentQueryTest.java
```diff
@@ -475,6 +476,73 @@ void withCondition_predicateMatched_getsComponents() {
...
+    void withText_null_getsAllComponent() {
+        TextComponent span1 = new TextComponent("sample text");
+        TextComponent span2 = new TextComponent("other text");
+
+        UI.getCurrent().getElement().appendChild(span1.getElement(),
+                span2.getElement());
+
+        Assertions.assertIterableEquals(List.of(span1, span2),
+                $(TextComponent.class).withText(null).allComponents());
+    }
```
caalador, line 511: This is perhaps a non test as it tests that if we don't set anything everything works as by default.

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/ElementConditionsTest.java
```diff
@@ -0,0 +1,142 @@
...
+        public HasTextComponent(String text) {
+            super(text);
+        }
+
+        @Override
+        public String getText() {
+            return HasText.super.getText().toUpperCase();
+        }
+    }
+}
```
caalador, line 142: new line

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/ComponentQueryTest.java
mcollovati: removed test

vaadin-testbench-core/src/test/java/com/vaadin/testbench/unit/ElementConditionsTest.java
mcollovati: fixed

caalador (approved):

PR #1392, opened 2022-05-24
refactor: move UI Unit testing to own module

caalador:
Mode the ui unit test code to its own
module and add the dependency to
testbench bom.


vaadin-testbench-unit/pom.xml
```diff
@@ -0,0 +1,412 @@
...
+                                    <!-- Package format version - do not
+                                        change -->
+                                    <Vaadin-Package-Version>1</Vaadin-Package-Version>
+                                </manifestEntries>
+                            </archive>
+                            <excludes>
+                                <!-- Remove resources included due to gwt
+                                    compilation -->
+                                <exclude>**/demoandtestapp/*</exclude>
+                            </excludes>
```
mcollovati, line 105: Is this needed also for UI unit test module?

vaadin-testbench-unit/pom.xml
```diff
@@ -0,0 +1,412 @@
...
+            <version>1.15</version>
+        </dependency>
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-remote-driver</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.seleniumhq.selenium</groupId>
+            <artifactId>selenium-java</artifactId>
+        </dependency>
```
mcollovati, line 316: Can we remove this dependencies from this module?

mcollovati (changes requested):

vaadin-testbench-unit/pom.xml
```diff
@@ -0,0 +1,412 @@
...
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>license-checker</artifactId>
+            <version>1.2.1</version>
+        </dependency>
+        <dependency>
+            <groupId>com.vaadin.external.gwt</groupId>
+            <artifactId>gwt-elemental</artifactId>
+            <version>2.8.2.vaadin2</version>
+        </dependency>
```
mcollovati, line 374: Can we remove this dependencies from this module?

vaadin-testbench-unit/pom.xml
caalador: removed

vaadin-testbench-unit/pom.xml
caalador: removed extra dependencies

vaadin-testbench-unit/pom.xml
caalador: removed

caalador: javassist perhaps not, but license-checker should be there as this is part of TestBench so it should check the TB license when running.

mcollovati: Ok, so maybe it will make sense to introduce a property on parent pom for `license-checker` version, so that it is defined in a single .
But we can do it later

mcollovati (approved):

PR #1386, opened 2022-05-21
feat: added mixin interfaces for wrappers

mcollovati:
Wrapper mixin interfaces provides specific overloads of method in order
to reduce the number of explicit casts of wrapper references in unit tests.

Closes #1379

PR #1395, opened 2022-05-24
chore: Make ComponentQuery generics friendly

mcollovati:
ComponentQuery terminal operators returns wrappers and components for
the exactly generic type. This is a little bit annoying on consumer perspective
when the component is generic itself.
For example with `GridWrap<Grid<Bean>,Bean> grid_ = $(Grid.class).first()`
compiler complains about Grid not being a Grid<Bean>.
This patch uses wildcards and explicit casts on terminal operators, to
make client API more friendly, so that
`GridWrap<Grid<Bean>,Bean> grid_ = $(Grid.class).first()`
works without need for casts.

### DIFF
--- a/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/ComponentQuery.java
+++ b/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/ComponentQuery.java
@@ -487,7 +487,7 @@ public class ComponentQuery<T extends Component> {
      *             if no component is found
      */
     @SuppressWarnings("unchecked")
-    public <X extends ComponentWrap<T>> X first() {
+    public <X extends ComponentWrap<? extends T>> X first() {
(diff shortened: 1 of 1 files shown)


knoobie:
Just thinking out loud, instead of 

```java

GridWrap<Grid<SamplePerson>, SamplePerson> grid_ = $(Grid.class)
                .first();

```

What would you think of: 

```java

var grid_ = $(Grid.class, SamplePerson.class)
                .first();

```

or 

```java

var grid_ = $(Grid.class).withType(SamplePerson.class)
                .first();

```


because people often use the var from java 11 and above, making the search a little bit more readable in my opinion

mcollovati:
Good suggestion, it needs to be investigated.
At a first glance such a feature may require having wrapper type as generic parameter for `ComponentQuery` and overloads of `$(componentType, itemType)` for every generic component (something similar to what is in the mixin interface for wrappers PR #1386)
Do you mind open a new issue for this?

PR #1403, opened 2022-05-25
feat!: Use test class package as base for classpath scan

mcollovati:
Restrict by default classpath scanning to current test class package and
subpackages.
Subclasses with different needs will overwrite scanPackage() method.

### DIFF
--- a/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -36,8 +36,9 @@ import com.vaadin.flow.router.RouteParameters;
  *
  * Provides methods to set up and clean a mocked Vaadin environment.
  *
- * Subclasses should typically restrict classpath scanning to a specific package
- * for faster bootstrap, by overriding {@link #scanPackage()} method.
+ * Classpath scanning is restricted to test class package and its subpackages
+ * for faster bootstrap, but surface can be widened or narrowed by overriding
+ * {@link #scanPackage()} method.
  *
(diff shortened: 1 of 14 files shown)


vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -111,10 +112,12 @@ protected void cleanVaadinEnvironment() {
      * Provide {@literal null} or empty string to scan the whole classpath, but
      * note that this may be quite slow.
      *
+     * Default is to scan the test class package and subpackages.
+     *
      * @return package name for classpath scanning.
      */
     protected String scanPackage() {
-        return null;
+        return getClass().getPackageName();
```
caalador, line 120: This change will actually break the start.vaadin.com view tests as the main layout references in the menu all view classes and the RouterLink will fail if they are not in the registry.
(This was not a problem with the master-detail branch as it only had the one view)

Also we should perhaps in the case of no routes found for a package log a warning that there were no registered routes found for the scanned package so that the developer could see that the scanPackage should be set.

Also the route missing error is quite confusing as it depends on the package scanned.

mcollovati: Very good point. Effectively do not automatically restrict classpath scanning seems the best option, because having main layouts and links to view in other packages may be a common use case.
I think this PR can be closed.
I'll just pick and adapt the second commit in another PR, in order to have a custom view for every wrapper

mcollovati:
Closing, as this PR introduces more troubles than benefits.
Test view per wrapper will be picked in a separate PR

PR #1412, opened 2022-05-30
feat: enable custom wrapper locations

caalador:
Add annotation Scan that can be given
an array of package strings that
will be scanned for ComponentWrappers.

This enabled custom wrappers outside
of the com.vaadin.testbench.component
package.


vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
```diff
@@ -0,0 +1,22 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Scan {
+
+    String[] value();
```
mcollovati, line 21: Can we add a javadoc to explain that `value` should be an array of package names?

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
```diff
@@ -0,0 +1,22 @@
...
+ * For the full License, see <https://vaadin.com/license/cvdl-4.0>.
+ */
+package com.vaadin.testbench.unit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
```
mcollovati, line 17: Do we support method level annotation?

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
```diff
@@ -0,0 +1,22 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Scan {
```
knoobie, line 19: The name could be anything or nothing.. could this be more elaborative like `@TestBenchUIScan` if it meant to be used by developer to specify their custom wrapper packages.

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
caalador: No that was a test, but it didn't really work. Removed

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
caalador: Added javadocs

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/Scan.java
caalador: TestBenchUIScan sounds like it scans for UI implementations.
Renamed to `@ComponentWrapPackages`

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/ComponentWrapPackages.java
```diff
@@ -0,0 +1,37 @@
...
+package com.vaadin.testbench.unit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to use to scan given packages for component wrappers outside the
+ * default "com.vaadin.flow.component`.
```
mcollovati, line 19: minor: package has double quote as prefix and backtick as suffix. Maybe we can use `@code` or `@literal` to wrap the package name.
```suggestion
 * default {@code com.vaadin.flow.component}.
```

mcollovati (approved):

PR #1417, opened 2022-05-31
feat: removing modal component creates roundtrip

### LINKED ISSUE #1349: Extends setting modality in tests
When making a component modal the inert state is only updated when the changes are written to the client side.
This means that in UIUnit testing just calling `UI.getCurrent().setChildComponentModal(span, true);` will not make it modal and other components inert as you need to fake a client change with 
`UI.getCurrent().getInternals().getStateTree().collectChanges(nodeChange -> {});`

This would be nice to have automatic so that you don't need to spend time trying to figure out that you need to make a "roundtrip" for the modality to work.

caalador:
An set modal component now gets a
detach listener that does a roundtrip
to propagate the modality change.

Fixes #1349

### DIFF
--- a/vaadin-testbench-unit/src/main/kotlin/com/vaadin/testbench/unit/mocks/MockedUI.kt
+++ b/vaadin-testbench-unit/src/main/kotlin/com/vaadin/testbench/unit/mocks/MockedUI.kt
@@ -9,10 +9,27 @@
  */
 package com.vaadin.testbench.unit.mocks
 
+import com.vaadin.flow.component.Component
+import com.vaadin.flow.component.ComponentEventListener
 import com.vaadin.flow.component.UI
+import com.vaadin.flow.shared.Registration
+import java.util.concurrent.atomic.AtomicReference
 
 /**
  * A simple no-op UI used by default by [com.vaadin.testbench.unit.MockVaadin.setup].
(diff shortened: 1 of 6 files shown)


vaadin-testbench-unit/src/main/kotlin/com/vaadin/testbench/unit/mocks/MockedUI.kt
```diff
@@ -9,10 +9,27 @@
...
  * A simple no-op UI used by default by [com.vaadin.testbench.unit.MockVaadin.setup].
  * The class is open, in order to be extensible in user's library
  */
-open class MockedUI : UI()
\ No newline at end of file
+open class MockedUI : UI() {
+
+    override fun setChildComponentModal(childComponent: Component?, modal: Boolean) {
+        super.setChildComponentModal(childComponent, modal)
+        if (modal) {
```
mcollovati, line 26: Is a round trip not needed when child is set to be modeless?

caalador: makes sense to execute also for that. Added roundtrip and removed it from ComponentWrap::setModal

PR #1432, opened 2022-06-02
feat: Add wrap for NumberField and TextArea

caalador:
Also update TextFieldWrapTest to
use a registered view instead of trusting
an external view to be available.

part of #1371

PR #1436, opened 2022-06-03
test: Create specific test view for notification component wrapper

mcollovati:


mcollovati:
Should be rebased after merging #1432 that already moved the view for TextFieldWrap

caalador (approved):

PR #1439, opened 2022-06-03
refactor!:view scan restriction by annotation

### LINKED ISSUE #1420: Extend package scanning so definition can be made by class
Defining package to scan for routes should be extended so that the package(s) to scan could be given as a class(es) (for instance the view class) instead of a string.

Also enabling multiple packages might make sense so you could pinpoint locations to scan.

Lets remove the scanPackages method and add a ViewPackages annotation that has Class[] values and String[] packages.

If the annotation is not present then use "" for scanning else use the defined class packages and string packages for scanning for routes.

Should we have also a boolean for do not scan?

mcollovati:
Introduced the ViewPackages annotation for classpath scan restriction.
It replaces the scanPackage method, that has been removed.

Fixes #1420

### DIFF
--- a/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -57,7 +57,7 @@ import com.vaadin.testbench.unit.mocks.MockedUI;
  */
 class BaseUIUnitTest {
 
-    private static final ConcurrentHashMap<Set<String>, Routes> routesCache = new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<String, Routes> routesCache = new ConcurrentHashMap<>();
 
     protected static final Map<Class<?>, Class<? extends ComponentWrap>> wrappers = new HashMap<>();
     protected static final Set<String> scanned = new HashSet<>();
@@ -125,9 +125,14 @@ class BaseUIUnitTest {
     }
 
     static synchronized Routes discoverRoutes(Set<String> packageNames) {
-        packageNames = packageNames == null ? Set.of() : packageNames;
-        return routesCache.computeIfAbsent(packageNames, pn -> new Routes()
-                .autoDiscoverViews(pn.toArray(String[]::new)));
+        packageNames = packageNames == null || packageNames.isEmpty()
+                ? Set.of("")
+                : packageNames;
+
+        return packageNames.stream()
+                .map(pkg -> routesCache.computeIfAbsent(pkg,
+                        p -> new Routes().autoDiscoverViews(p)))
+                .reduce(new Routes(), Routes::merge);
     }
 
     protected void initVaadinEnvironment() {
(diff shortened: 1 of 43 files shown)


vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -113,16 +120,19 @@ private static Map<Class<?>, Class<? extends ComponentWrap>> scanWrappers(
...
-        return routesCache.computeIfAbsent(packageName,
-                pn -> new Routes().autoDiscoverViews(pn));
+    synchronized Routes discoverRoutes() {
+        return discoverRoutes(scanPackages());
+    }
+
+    static synchronized Routes discoverRoutes(Set<String> packageNames) {
+        packageNames = packageNames == null ? Set.of() : packageNames;
+        return routesCache.computeIfAbsent(packageNames, pn -> new Routes()
+                .autoDiscoverViews(pn.toArray(String[]::new)));
```
caalador, line 130: Instead of changing the hash to depend on a set we should just do the compute if absent for each packagename and add build up a Routes object from each of them.
Else the caching is a bit too specific to be of use for multiple tests with different packages.

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -137,6 +147,23 @@ void scanForWrappers() {
         }
     }
 
+    Set<String> scanPackages() {
+        Set<String> packagesToScan = new HashSet<>();
+        ViewPackages packages = getClass().getAnnotation(ViewPackages.class);
+        if (packages != null) {
+            Stream.of(packages.classes()).map(Class::getPackageName)
```
caalador, line 154:
```suggestion
        if (getClass().isAnnotationPresent(ViewPackages.class)) {
            ViewPackages packages = getClass().getAnnotation(ViewPackages.class);
            Stream.of(packages.classes()).map(Class::getPackageName)
```

vaadin-testbench-unit/src/test/java/com/vaadin/flow/component/accordion/AccordionWrapTest.java
```diff
@@ -15,14 +15,11 @@
 
 import com.vaadin.flow.router.RouteConfiguration;
 import com.vaadin.testbench.unit.UIUnitTest;
+import com.vaadin.testbench.unit.ViewPackages;
 
+@ViewPackages(packages = "com.vaadin.flow.component.accordion")
```
caalador, line 20: Could be without a defined package. The test was just written before choosing to use the getClass().getPackageName() which is simpler to write and understand.

caalador (changes requested):

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
mcollovati: Changed by adding a `merge` operation to `Routes` that makes a copy with data from two other `Routes`.
Did not use the constructor directly because it takes a mutable sets and passing the ones from the source route may corrupt them with unwanted data.

vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
mcollovati: fixed

vaadin-testbench-unit/src/test/java/com/vaadin/flow/component/accordion/AccordionWrapTest.java
mcollovati: fixed

caalador (approved):

PR #1477, opened 2022-06-14
feat: Added query for component method to ComponentWrap

mcollovati:
Allows to query for a nested component directly from
the ComponentWrap instance

Fixes #1445


vaadin-testbench-unit/src/main/java/com/vaadin/testbench/unit/ComponentWrap.java
```diff
@@ -269,4 +271,67 @@ protected void fireDomEvent(DomEvent event) {
...
+     * @param componentType
+     *            the type of the component to search for
+     * @param queryBuilder
+     *            the function that sets query condition
+     * @param <R>
+     *            the type of the component to search for
+     * @return the component found by query execution, wrapped into an
+     *         {@link Optional}, or empty if the query does not produce results.
+     */
+    protected <R extends Component> Optional<R> findByQuery(
```
caalador, line 294: Why are the methods protected?
Also would it be better if this was instead just returning a pre-populated from ComponentQuery?
e.g. `public <R extends Component> ComponentQuery<R> find(Class<R> componentType)` from which you can then continue like if you would have started a normal search.

mcollovati: I was thinking this method just as an internal helper for sub-classes.
Passing the query to a consumer was just to be able to enforce `from` to the current component.
I'll refactor as you suggest

mcollovati: Added the `find` method, but left the `findByQuery` and `findAllByQuery` as helper for subclasses.
Can be remove in the future if not needed.

PR #1479, opened 2022-06-15
feat: add method to get the item component

### LINKED ISSUE #1467: A way to get the component for a grid cell
There seems to be only `getCellText(row, column)` but in the case I have used a `ComponentRenderer` to add a component to a cell, I would want to be able to access that. For instance if I add a button to a cell, I want to be able to click it. If I add an "edit" link, I might want to assert that it points to the correct URL

caalador:
Add method for getting the component
for a cell in a grid.

Fixes #1467

### DIFF
--- a/vaadin-testbench-unit/src/main/java/com/vaadin/flow/component/grid/GridWrap.java
+++ b/vaadin-testbench-unit/src/main/java/com/vaadin/flow/component/grid/GridWrap.java
@@ -573,14 +573,4 @@ public class GridWrap<T extends Grid<Y>, Y> extends ComponentWrap<T> {
         }
     }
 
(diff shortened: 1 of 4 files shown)


vaadin-testbench-unit/src/main/java/com/vaadin/flow/component/grid/GridWrap.java
```diff
@@ -521,4 +573,14 @@ private String getColumnInternalId(Grid.Column targetColumn) {
         }
     }
 
+    private String getColumnKey(Grid.Column targetColumn) {
```
mcollovati, line 576: Can't find where this method is used. Is it here as a helper for other features?

mcollovati (approved):

PR #1481, opened 2022-06-15
test: fixed wrong test

mcollovati:
Exception was thrown, but not for the expected reason, but because menu was not open.

### DIFF
--- a/vaadin-testbench-unit/src/test/java/com/vaadin/flow/component/contextmenu/ContextMenuWrapTest.java
+++ b/vaadin-testbench-unit/src/test/java/com/vaadin/flow/component/contextmenu/ContextMenuWrapTest.java
@@ -200,8 +200,10 @@ class ContextMenuWrapTest extends UIUnitTest {
     @Test
     void clickItem_byNestedIndexNotUsableParent_throws() {
         // Hierarchical / NestedDisabled / Level3
+        menu_.open();
+
         Assertions.assertThrows(IllegalStateException.class,
-                () -> menu_.clickItem(7, 2, 0));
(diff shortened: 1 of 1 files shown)


vaadin-testbench-unit/src/test/java/com/vaadin/flow/component/contextmenu/ContextMenuWrapTest.java
```diff
@@ -200,8 +200,10 @@ void clickItem_byInvalidIndexes_throws() {
     @Test
     void clickItem_byNestedIndexNotUsableParent_throws() {
         // Hierarchical / NestedDisabled / Level3
+        menu_.open();
+
         Assertions.assertThrows(IllegalStateException.class,
-                () -> menu_.clickItem(7, 2, 0));
+                () -> menu_.clickItem(7, 3, 0));
```
caalador, line 206: Should we for these actually have
```
IllegalStateException exception = Assertions.assertThrows(IllegalStateException.class,
                () -> menu_.clickItem(7, 3, 0));
Assertions.assertTrue(exception.getMessage().contains("THE EXPECTED MESSAGE"));
```
Just to make certain that the failure reason is correct as we do throw IllegalStateExceptions from many places.

mcollovati: Good idea. I'll add the assertion

mcollovati: Indeed there was another wrong test

caalador (approved):

PR #1489, opened 2022-06-16
feat!: change wrap naming

### LINKED ISSUE #1488: Change Wrap to Tester and wrap(component) to test(component)

caalador:
Change Wrap to Tester and
the wrap method to test.
Also change Wraps annotation to Tests

fixes #1488


mcollovati:
Should we rename also `@ComponentWrapPackages` to `@ComponentTesterPackages`?

mcollovati:
We should maybe rename also test classes (eg `RadioButtonGroupWrapTest` -> `RadioButtonGroupTesterTest`) but this can also be done in a separate PR.
Also some javadocs may be updated to use the term `tester` instead of `wrapper`, but again we can do it in another PR.

caalador:
Some javaDocs might still have the wrong naming. 
Some have been left at  `test wrapper` as it is not for the name, but for the usage.

caalador:
Thanks, way too many files changing all around the project

mcollovati (approved):

PR #1495, opened 2022-06-21
refactor: make MockRouteNotFoundError extend RouteNotFoundError

mcollovati:
Making MockRouteNotFoundError extending RouteNotFoundError and
annotating it with `@DefaultErrorHandler` won't make Vaadin apps fail at
startup when vaadin-testbench-unit incidentally is on classpath and
would also preserve the output from the out-of-the-box handler, so
testbench tests checking for it would continue to work without any
change.


mcollovati:
Tested locally against flow test-root-context ITs.
Also running full test suite locally at the moment; will post results when completed.

caalador:
You need to test one of the theme tests as the overriding default handler will be used instead so the expectation for a file not found will fail as the shown error page will be the mock route not the error route.
Also what happens when there are 2 default routes for a single exception?

caalador:
As it actually returns the same content now it should be fine.

mcollovati:
Themes tests are green locally

mcollovati:
Yes, sorry. The comment is wrong. MockRouteNotFoundError is picked

caalador (approved):

PR #1504, opened 2022-07-06
fix!: Make Saucelabs integration use what Saucelabs recommends

Artur-:
Read tunnel identifier from sauce.tunnelId property or SAUCE_TUNNEL_ID env variable

Note that this changes the default URL used for Sauce Labs from http://localhost:4445/wd/hub to the recommended https://ondemand.saucelabs.com/wd/hub. If you have an existing setup relying on the old Maven plugin, you might need to override `getHubUrl` to return the old URL


alvarezguille:
Don't forget to update docs to remove sauce-connect-plugin mentions and explain what to do instead.

Artur-:
https://github.com/vaadin/docs/pull/1545

alvarezguille (approved):

PR #1508, opened 2022-07-08
fix: Avoid invalid Selenium deprecation warnings for Sauce parameters

Artur-:
Uses a map instead of DesiredCapabilities for the Sauce Labs options as Selenium apparently checks all DesiredCapabilities for unknown keys and not only the top level


Artur-:
Well this does not work. When duplicating `DesiredCapabilities` the map is not duplicated so the same Sauce labs options will be shared across all test methods. So all test methods will be named the same.

Artur-:
But actually this is already broken when using `DesiredCapabilities` so this PR does not make it any worse...

alvarezguille (approved):

Issue #1511, opened 2022-07-13
Allow configuring Hub URL when using saucelabs


alvarezguille:
Different users might have different endpoints and it should be possible to configure which one to use.
https://docs.saucelabs.com/basics/data-center-endpoints/#data-center-endpoints

PR #1512, opened 2022-07-13
feat: allow setting sauce hub url with property or env

alvarezguille:
Different users might have different saucelabs endpoints.
Now it's possible to configure it by using the `sauce.hubUrl` system property or the `SAUCE_HUB_URL` environment variable.
Possible values are available in https://docs.saucelabs.com/basics/data-center-endpoints/#data-center-endpoints

fixes: #1511


vaadin-testbench-core/src/main/java/com/vaadin/testbench/parallel/SauceLabsIntegration.java
```diff
@@ -173,7 +175,11 @@ private static String getTunnelIdentifierFromOptions(String options) {
      * @return url String to be used in Sauce Labs test run
      */
     static String getHubUrl() {
-        return "https://ondemand.us-west-1.saucelabs.com/wd/hub";
+        String hubUrl = getSystemPropertyOrEnv(SAUCE_HUB_URL_PROP, SAUCE_HUB_URL_ENV);
+        if (hubUrl == null) {
+            hubUrl = "https://ondemand.us-west-1.saucelabs.com/wd/hub";
```
mcollovati, line 180: nit: we can perhaps introduce a constant for the default hub URL

mcollovati (approved):

PR #1528, opened 2022-09-29
chore: Upgrade to Selenium 4.5.0

Artur-:


alvarezguille (changes requested):
could you bump the TB minor version as well? selenium 4.5 seems to have some deprecated API removals that could be unexpected in a maintenance release.
apart from this change we could have a separate PR for 8.1.x adding the latest `selenium-devtools-###`

Artur-:
bumped minor

Artur-:
Done in #1530

knoobie:
@alvarezguille Are there any plans that testbench is managing `selenium-devtools-v***` or should this be documented that developer should take care of it?

alvarezguille:
We had some internal discussions about how to handle selenium versions within patch releases now that it also affects chrome compatibility.
We currently have selenium 4.5.3 in TB 8.1.3 which is what this PR intended before my comment bump the minor. 
I'm not sure where in docs this could be mentioned but it's worth mentioning.
The intention is to keep updating selenium version as long as possible to avoid extra dependencies that were only released with selenium's previous version number. @artur- can you confirm?

PR #1530, opened 2022-10-03
chore: Upgrade to Selenium 4.5.0

alvarezguille:


alvarezguille:
After offline discussion decided with @Artur- not to bump minor version as in #1528 since adding support for newer selenium in previous minor can be troublesome as shown in #1529.

Artur- (approved):

PR #1531, opened 2022-10-05
chore: Update to LC 1.10.0

alvarezguille:
Use build type null in license check


ZheSun88:
there is an error in the build
```
error was: 400: Your Sauce Connect version (4.4.12) is no longer supported (must be newer than 4.5.4). Please download the latest version from https://docs.saucelabs.com/secure-connections/sauce-connect/installation (ref: ff65e012). HTTP status: 400 Bad Request
```

Artur- (approved):

PR #1536, opened 2022-10-10
feat: Support setting the read timeout for remote web drivers (#1518)

alvarezguille:
This allows you to increase the timeout e.g. when using Saucelabs so that tests wait for a session to become available instead of timing out and causing test failures


Artur- (approved):

alvarezguille:
This feature is not compatible with Selenium 3.x

alvarezguille:
Just for reference, the feature isn't really needed with selenium 3.x either https://github.com/SeleniumHQ/selenium/blob/selenium-3.141.59/java/client/src/org/openqa/selenium/remote/http/HttpClient.java#L94

PR #1572, opened 2022-11-15
test: update after dx tests

MarcinVaadin:
Changes after DX tests:

1. Moved parallel configuration to testbench-core-junit5 and enabled by default
2. Added custom parallel configuration to support testbench Parameters
3. Added driver closing on test failure while using screenshot extension

### DIFF
--- a/vaadin-testbench-core-junit5/pom.xml
+++ b/vaadin-testbench-core-junit5/pom.xml
@@ -32,7 +32,6 @@
 
     <properties>
         <junit5.version>5.9.1</junit5.version>
-        <junit5.platform.version>1.9.1</junit5.platform.version>
         <mockito.version>4.8.1</mockito.version>
     </properties>
 
@@ -178,6 +177,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>${junit5.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
@@ -209,17 +215,14 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
-            <version>${junit5.version}</version>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
-            <version>${junit5.version}</version>
         </dependency>
         <dependency>
             <groupId>org.junit.platform</groupId>
             <artifactId>junit-platform-engine</artifactId>
-            <version>${junit5.platform.version}</version>
         </dependency>
 
         <dependency>
--- a/vaadin-testbench-core-junit5/pom.xml
+++ b/vaadin-testbench-core-junit5/pom.xml
@@ -32,6 +32,7 @@
 
     <properties>
         <junit5.version>5.9.1</junit5.version>
+        <junit5.platform.version>1.9.1</junit5.platform.version>
         <mockito.version>4.8.1</mockito.version>
     </properties>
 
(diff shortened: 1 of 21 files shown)


vaadin-testbench-core-junit5/src/main/java/com/vaadin/testbench/BrowserTestBase.java
```diff
@@ -20,6 +22,7 @@
  * A superclass with helper methods to aid TestBench developers create a JUnit 5
  * based tests.
  */
+@Execution(ExecutionMode.CONCURRENT)
```
mshabarov, line 25: Hope this can be overriden in the test class?

vaadin-testbench-core-junit5/src/main/java/com/vaadin/testbench/parallel/ParallelConfigurationStrategy.java
```diff
@@ -0,0 +1,52 @@
...
+ * Custom configuration strategy using TestBench
+ * {@link Parameters#getTestsInParallel()}.
+ */
+public class ParallelConfigurationStrategy implements
+        ParallelExecutionConfiguration, ParallelExecutionConfigurationStrategy {
+
+    private static final int MAX_CONCURRENT_TESTS;
+
+    static {
+        MAX_CONCURRENT_TESTS = Parameters.getTestsInParallel();
```
mshabarov, line 19: What is the default value?

vaadin-testbench-core-junit5/pom.xml
```diff
@@ -32,6 +32,7 @@
 
     <properties>
         <junit5.version>5.9.1</junit5.version>
+        <junit5.platform.version>1.9.1</junit5.platform.version>
```
knoobie, line 35: If you import the junit5-bom the second property is not needed and correctly managed by junit5

vaadin-testbench-core-junit5/src/main/java/com/vaadin/testbench/parallel/ParallelConfigurationStrategy.java
MarcinVaadin: From `Parameters`:
```
if (hasSystemProperty("testsInParallel")) {
    testsInParallel = getSystemPropertyInt("testsInParallel", 1);
} else if (isLocalWebDriverUsed()) {
    testsInParallel = 10;
} else {
    testsInParallel = 50;
}
```

vaadin-testbench-core-junit5/src/main/java/com/vaadin/testbench/BrowserTestBase.java
MarcinVaadin: Yes, it can be overriden by `@Execution(ExecutionMode.SAME_THREAD)` or same result will be after setting `testInParallel` property value to 1.

vaadin-testbench-core-junit5/pom.xml
MarcinVaadin: Good point, updated. The bom name for junit5 is `junit-bom` - bit confusing.

vaadin-testbench-core-junit5/src/main/java/com/vaadin/testbench/parallel/ParallelConfigurationStrategy.java
mshabarov: So according to this code, since we don't have "testsInParallel" in default config, the execution would be parallelised into 10 or 50 threads. I did'n see this description in the docs, could you add a sentence there explaining this default behaviour and tell how to override?

MarcinVaadin: > Up to 50 test methods are executed simultaneously by default. The limit can be set using the com.vaadin.testbench.Parameters.testsInParallel system property.

Isn't it enough?

mshabarov: Yes, and it's worth to tell that if one want to disable concurrent runs without digging into parameters, he can use `@Execution(ExecutionMode.SAME_THREAD)`.

MarcinVaadin: Docs updated https://github.com/vaadin/docs/pull/1865/commits/c6751c6b872b2eff1aa5d7ecca0f76162e718035

mshabarov (approved):

PR #1596, opened 2023-01-20
test: fixed 172 private ip address range

mcollovati:
Not only "172.16", but a wider portion of the "172" address range, is designated for private use.

172.16.0.0/12 IP addresses: 172.16.0.0 – 172.31.255.255


Artur- (approved):
Regexps are horrible

PR #1602, opened 2023-02-10
chore: Add selenium-chromium-driver to testbench bom

ZheSun88:
this should be picked to 8.2 and 8.1 which are using selenium 4.0+, as this module has been introduced since Selenium 4.0


mshabarov (approved):
@ZheSun88 approved, but could you please give a context why is this needed? Does it fix some issue with TB?

Artur-:
This is exactly what I need to override the selenium versions in flow...

PR #1616, opened 2023-03-15
As the ShortcutRegistration will always use HashableKeys, hencewhy...

czp13:
… need to use it in our tests as well to do not have any differences

within our tests and real life application. (now we used simple Key objects, but we need to use HashableKeys)

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.


vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
```diff
@@ -86,8 +115,13 @@ public fun Component._fireShortcut(key: Key, vararg modifiers: Key) {
...
     // as if it contained all filters (otherwise `matchesFilter()` would NPE on missing key)
     // and respond true only to the matching filter.
+
+    // we need to use hashableKeys not simple keys, as this is the implementation used in ShortcutRegistration
+    // class.
+    val hashableKeys = modifiers.map { getHashableKey(it) }
+
     @Suppress("UNCHECKED_CAST")
-    val data = MockFilterJsonObject(key, modifiers.toSet() as Set<KeyModifier>)
+    val data = MockFilterJsonObject(key, hashableKeys.toSet() as Set<KeyModifier>)
```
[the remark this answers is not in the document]
czp13: Yes, Marco, I shared on Slack that the mentioned line changed, so my Monday/yd changes caused the difference between the testbench and flow, how they call this private method, which has triggered the test failures. I can definitely rewrite the flow part to help address the issue.

However, I would emphasize that generateEventModifierFilter() is always called with HashableKeys right now:
(and some other code parts making sure it)
```
    private void addKey(Key key) {
        assert key != null;

        HashableKey hashableKey = new HashableKey(key);

        if (Key.isModifier(key)) {
            if (!modifiers.contains(hashableKey)) {
                modifiers.add(hashableKey);
                prepareForClientResponse();
            }
        } else {
            if (primaryKey == null || !primaryKey.equals(hashableKey)) {
                primaryKey = hashableKey;
                prepareForClientResponse();
            }
        }
    }
```
Since generateEventModifierFilter() is a private method that is called through other methods and in a state when HashableKeys have already been added to the ShortcutRegistration object, it would make sense to call it properly with HashableKeys, just like in any other use case, it is not an open API it shall not be called with Key objects, and potentially method signature could have been changed as well to be less flexible and refer to this.

So another solution would be to make the flow side (the private method) more rigid (so use HashableKeys in the method contract, and parameters), to catch errors even before runtime, as this is a specific, exact use case for this private method.

I feel like we want to force the Liskov principle to private methods :D. 

**But as the consensus here as I feel is to make the methods more flexible, with less constraint, I can do that, let me make the PR for it.**

Worth mentioning that by doing this we can have more unexpected behavior in runtime as we are more flexible. 
But I am doing this just to share my opinion and ideas about the optimal solution.

I need a bit of time again to build the flow artifact, include it in a customer project, and manually as well trying out many combinations to make sure it is working properly (I always do manual tests too).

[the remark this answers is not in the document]
czp13: So my TLDR again:D 

I think in general it makes sense for private methods to have them more rigid, and less flexible/accepting for input params to ensure correct behavior and catch incorrect working in compile time. My two cents! 🙇

vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
```diff
@@ -61,6 +65,34 @@ private class MockFilterJsonObject(val key: Key, val modifiers: Set<KeyModifier>
...
+    IllegalAccessException::class
+)
+public fun getHashableKey(keyModifier: Key): Any? {
+    val nestedClass = Class.forName(
+        "com.vaadin.flow.component.ShortcutRegistration\$HashableKey"
+    )
+    val ctor = nestedClass.declaredConstructors[0]
+    ctor.isAccessible = true
+    return ctor.newInstance(keyModifier)
+}
```
mcollovati, line 94: Does this method need to be public? Or can we move it into the `MockFilterJsonObject` class or in the companion object and make it private or internal?
It seems to me it is not used elsewhere

vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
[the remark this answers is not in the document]
czp13: I thought I replied to this, sorry. 

Sure we can do this, I already did a commit containing the changes to do the HashAbleKey transformation within the MockFilterJsonObject! 🙇

vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
czp13: Yes, we can do this as well, makes sense as we only use it from that class.
Also I got these errors:
``` java.lang.ClassCastException: class com.vaadin.flow.component.Key$1 cannot be cast to class com.vaadin.flow.component.KeyModifier (com.vaadin.flow.component.Key$1 and com.vaadin.flow.component.KeyModifier are in unnamed module of loader 'app')```

Hencewhy, I changed some casting and the class variable types. Anytime I tried to get back the Key from the Keymodifier set, I got this.... 

Some kind of anonymous class/Kotlin magic is happening here that Key$1 is generated, but I would spare my nerves and would not go further down the road and I just fix this with the proper typing/class names in the methods... (I just wanted initially to have this strict as well, but we do enough magic here (reflection for private method, inner class constructor, emulating filterText, etc)...

I will commit your suggestion and just leaving as it is this change:

File: vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt (hunk 1 above)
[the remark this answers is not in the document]
Comment (czp13) [reply vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt]: Done ✅. 
Please approve if you feel, have time and do not have any further request, comment. :)

File: vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt (hunk 2 above)
Comment (czp13) [reply vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt]: The method replacement is Done ✅.
Please approve if you feel, have time and do not have any further requests, comments. :)

File: vaadin-testbench-unit-shared/src/main/java/com/vaadin/flow/component/grid/GridTester.java [hunk 3]
```diff
@@ -560,8 +560,7 @@ private String getValueProviderString(int row, Grid.Column targetColumn)
             ColumnPathRenderer renderer = (ColumnPathRenderer) targetColumn
                     .getRenderer();
 
-            Field f = ColumnPathRenderer.class
-                    .getDeclaredField("provider");
+            Field f = ColumnPathRenderer.class.getDeclaredField("provider");
```
mcollovati, line 563: I would revert this formatting change, so that the pr only focus on the shortcut issue

mcollovati (changes requested):

vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
```diff
@@ -85,9 +81,13 @@ private class MockFilterJsonObject(val key: Key, val modifiers: Set<Key>) : JreJ
     companion object {
         private val mgenerateEventKeyFilter = ShortcutRegistration::class.java.getDeclaredMethod("generateEventKeyFilter", Key::class.java)
         private val mgenerateEventModifierFilter = ShortcutRegistration::class.java.getDeclaredMethod("generateEventModifierFilter", Collection::class.java)
+        private val chashableKey =  Class.forName(
+            "com.vaadin.flow.component.ShortcutRegistration\$HashableKey"
+        ).declaredConstructors[0]
```
czp13, line 86: very last thing is we could have this:
```
 ShortcutRegistration::class.java
    .classLoader
    .loadClass("com.vaadin.flow.component.ShortcutRegistration\$HashableKey")
    .declaredConstructors[0]
```
Not sure if it is better or not, maybe a bit more Kotlin syntax.

vaadin-testbench-unit-shared/src/main/java/com/vaadin/flow/component/grid/GridTester.java
czp13: 👍

czp13: I just pushed mvn formatter:format. I can revert it.

mcollovati (approved):

PR #1620, opened 2023-03-17
fix: changing Shortcuts.kts logic to use HashableKeys. (#1616) (#1619)

mshabarov:
* As the ShortcutRegistration will always use HashableKeys, hence why we need to use it in our tests as well to not have any differences between our tests and real-life application. (now we used simple Key objects, but we need to use HashableKeys)

* Cleaned up the code a bit to use the HashableKey creation only in the MockFilterJsonObject class. (Based on @mcollovati  review, thank you)

* Put the getHashableKey() method into the class (mock JSON), so where we use it (suggestion by @mcollovati, thanks!).

* One more time, even nicer this way, like this one :) (Thanks again @mcollovati)

* Revert GridTester changes, and fix java doc for the newly added getHashableKey() method, fix imports.

* Using a bit more Kotlinish syntax for reflection (private inner class constructor), and being more strict on that we are using ShortcutRegistration class. (discussed with @mcollovati that this could be a better way)

* Enabling Shortcut related tests (were disabled by another dependency FIX pr to merge it to master)

(cherry picked from commit baf3886f054e95cc1dd2fd54642db203fdb42d05)

### DIFF
--- a/vaadin-testbench-unit/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
+++ b/vaadin-testbench-unit/src/main/kotlin/com/vaadin/testbench/unit/internal/Shortcuts.kt
@@ -2,7 +2,6 @@ package com.vaadin.testbench.unit.internal
 
 import com.vaadin.flow.component.Component
 import com.vaadin.flow.component.Key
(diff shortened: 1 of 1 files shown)


czp13 (approved):
Looks good as well, thank you!

PR #1625, opened 2023-03-30
fix: Ignore spurious chromedriver error

Artur-:
At least fixes a navigation test in Flow

### DIFF
--- a/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/TestBenchElement.java
+++ b/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/TestBenchElement.java
@@ -177,6 +177,12 @@ public class TestBenchElement implements WrapsElement, WebElement, HasDriver,
             // Avoid strange "element not clickable at point" problems
             callFunction("click");
         } catch (Exception e) {
+            if (e.getMessage()
+                    .contains("Inspected target navigated or closed")) {
+                // This happens with chromedriver although e.g. navigation
+                // succeeds
+                return;
+            }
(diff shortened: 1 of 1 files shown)


Artur-:
See e.g. https://github.com/electron-userland/spectron/issues/493 for a similar issue

mcollovati:
It seems like someone fixed it by setting the `--disable-site-isolation-trials` property in chrome options, but I there's no explanation.

https://github.com/angular/protractor/issues/5347#issuecomment-686722070

https://stackoverflow.com/a/56528802

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/TestBenchElement.java
```diff
@@ -177,6 +177,12 @@ public void click() {
             // Avoid strange "element not clickable at point" problems
             callFunction("click");
         } catch (Exception e) {
+            if (e.getMessage()
+                    .contains("Inspected target navigated or closed")) {
+                // This happens with chromedriver although e.g. navigation
+                // succeeds
+                return;
```
mcollovati, line 184: Did I understand correctly that if this error occurs, the action is already completed correctly, and it is safe to ignore the exception?

Artur-: That is what happens in the flow tests. The navigation completes, then an exception is thrown and because of the logic in TestBenchElement.click, the action is retried and fails as navigation did happen

mcollovati (approved):

PR #1654, opened 2023-07-18
chore: Add missing NativeLabelTester

TatuLund:
In Vaadin 24.1 NativeLabel replaced Label


mcollovati:
The current testbench version is compatible with Vaadin 24.0, that however does not contain NativeLabel (from 24.1)

mshabarov:
This now fails with `Check contributor CI rights`.
@ZheSun88 do you know how to fix it?

PR #1661, opened 2023-08-10
fix: Make BaseUIUnitTest extendable

### LINKED ISSUE #1660: Expose BaseUIUnitTest in order to be easier to extend
Currently UI Unit Test feature class hierarchy is as follows.: UIUnitTest and SpringUIUnitTest for JUnit 5 and UIUnit4Test and SpringUIUnit4Test for older JUnit versions. These classes are all "simple" classes that inherit from BaseUIUnitTest and add JUnit support. 

Now consider extending UI Unit Test to be used in non-JUnit framework, namely the TestNG. This would be straightforward to do if one could inherit in the same way from BaseUIUnitTest but this time use the TestNG annotations. This does not work because BaseUIUnitTest is package private.

TatuLund:
Fixes #1660

### DIFF
--- a/vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -58,7 +58,7 @@ import com.vaadin.testbench.unit.mocks.MockedUI;
  *
  * @see ViewPackages
  */
-abstract class BaseUIUnitTest {
+public abstract class BaseUIUnitTest {
 
     private static final ConcurrentHashMap<String, Routes> routesCache = new ConcurrentHashMap<>();
 
@@ -127,7 +127,8 @@ abstract class BaseUIUnitTest {
         return discoverRoutes(scanPackages());
     }
 
-    static synchronized Routes discoverRoutes(Set<String> packageNames) {
+    protected static synchronized Routes discoverRoutes(
+            Set<String> packageNames) {
         packageNames = packageNames == null || packageNames.isEmpty()
                 ? Set.of("")
                 : packageNames;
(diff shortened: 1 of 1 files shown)


mshabarov (approved):

vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -143,7 +144,7 @@ protected void initVaadinEnvironment() {
         MockVaadin.setup(discoverRoutes(), MockedUI::new, lookupServices());
     }
 
-    void scanTesters() {
+    protected void scanTesters() {
```
mcollovati, line 147: Needs javadoc, as it is now exposed to end users

vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
mcollovati, line 1: Since we are making `BaseUIUnitTest` a public API, we should probably update the Javadoc to remove the `For internal use only` comment and to offered document the extension points (`initVaadinEnvironment`, `cleanVaadinEnvironment`, ...) and utility methods (`discoverRoutes`, `scandTesters`).

mcollovati (changes requested):

vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
mcollovati: Thinking about `For internal use only`, it is maybe better to leave it there, as we are not suggesting to extend this class, and we should be free to change internals even with breaking changes.

mcollovati (approved):

PR #1671, opened 2023-08-21
chore: Use Vaadin 24.2 and upgrade project version

mshabarov:


mshabarov:
This discovers a blocker for TestBench 9 to use Vaadin 24.2:
```
Compiler
.../vaadin-testbench-unit-shared/src/main/kotlin/com/vaadin/testbench/unit/internal/MockVaadin.kt:[227,33] Unresolved reference: SERVER_ROUTING
```
We have to change the implementation.

PR #1673, opened 2023-08-30
fix add selenium-devtools to 8.2

caalador:


vaadin-testbench-bom/pom.xml
```diff
@@ -113,6 +113,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>org.seleniumhq.selenium</groupId>
+                <artifactId>selenium-devtools-v115</artifactId>
+                <version>4.11.0</version>
```
mshabarov, line 119: This version differs from selenium version defined in a root pom - 4.8.3. Do we want to set it anyway?

PR #1683, opened 2023-09-27
Update Selenium to 4.13

alvarezguille:


vaadin-testbench-shared/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java
```diff
@@ -66,7 +66,9 @@ static public WebDriver createDriver(
                 FirefoxProfile profile = new FirefoxProfile(profileDir);
                 options.setProfile(profile);
             }
-            options.setHeadless(Parameters.isHeadless());
+            if (Parameters.isHeadless()) {
+                options.addArguments("-headless");
```
mshabarov, line 70: Maybe it's fine, as the validation is green, but shouldn't it be --headless (with two dashes).

alvarezguille: I took the argument from the deprecation comment
https://github.com/SeleniumHQ/selenium/commit/dbcad078e520c80605b4fce856ceda30be33fc93#diff-d011ae7f625544736a498af510308e81e40d69d9d85e7fee18c14aedc4ed4d9bL259

mshabarov (approved):

Issue #1635, opened 2023-05-05
Add Testers for SideNav, TabSheet and MultiSelectComboBox / UI Unit Tests


TatuLund:
New components, notably SideNav, TabSheet and MultiSelectComboBox seem to lack Testers.

czp13:
Thanks, Tatu. Shall be put here probably:
- https://github.com/vaadin/testbench/tree/main/vaadin-testbench-unit-shared/src/main/java/com/vaadin/flow/component

TatuLund:
@czp13 exactly.

PR #1691, opened 2023-10-25
feat: add tester for SideNav component

mcollovati:
Also refactors ComponentTester to expose some utility method to subclasses.

Part of #1635


mcollovati:
The tester can potentially be further improved by providing methods to access items by index (as done in `MenuBarTester`) and to get single items instance.
We can also consider if it is worth it adding a tester for `SideNavItem`.

caalador (approved):

czp13 (approved):

PR #1694, opened 2023-10-26
fix: make tabs tester index operations work on visible tabs

mcollovati:
The Tabs tester index base operation currently does not care about tabs visibility, thus not correctly reflecting the action in the browser. If there are three tabs, but the one in the middle is hidden, operation for index 2 should effectively act on the third registered tab. This change fixes fetching the tab by index by filtering out hidden tabs.


vaadin-testbench-unit-shared/src/main/java/com/vaadin/flow/component/tabs/TabsTester.java
```diff
@@ -54,8 +55,12 @@ public void select(String label) {
      */
     public void select(int index) {
```
caalador, line 56: I would perhaps update the JavaDoc to underline the case that non visible tabs are not calculated into the index when selecting.

caalador (approved):

Issue #1705, opened 2023-12-11
Wait for dev bundle build in waitForVaadin


mshabarov:
Vaadin Flow 24.3 has now a "Building front-end development bundle" spinner shown on the page when dev bundle is being generated, see [this issue](https://github.com/vaadin/flow/issues/16607).

TestBench should take this into account in `waitForVaadin` [method](https://github.com/vaadin/testbench/blob/06271bd9ca67c61e4a04a3dc07ee32bc771074c9/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java#L107) to let Flow apps to properly wait until custom dev bundle is ready.

Possible solution is to wait until spinner div 
```xml
<div class="message">Building front-end development bundle</div>
```

disappears, like shown below:

```java
waitUntil(driver -> {
            try {
                TestBenchElement testBenchElement = $("div").attribute("class", "message").waitForFirst();
                return testBenchElement == null;
            } catch (TimeoutException e){
                return true;
            } catch (NoSuchElementException e){
                return true;
            }
        });
```

TatuLund:
I have used slightly different wayt to do the same 

https://github.com/TatuLund/bean-table/blob/v24/src/test/java/org/vaadin/tatu/AbstractViewTest.java#L123

mshabarov:
UPD: TestBench `waitForVaadin` method already has a check for dev server [here](https://github.com/vaadin/testbench/blob/06271bd9ca67c61e4a04a3dc07ee32bc771074c9/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java#L60C44-L60C44).

This checks for a flag `devServerIsNotLoaded` which is added on "dev mode not ready" page [here](https://github.com/vaadin/flow/blob/c8edbf0c3a77cf57bc1af65af3df419c2defa149/vaadin-dev-server/src/main/resources/com/vaadin/base/devserver/dev-mode-not-ready.html#L52).

We should start with testing the `waitForVaadin()` method properly waiting until custom dev bundle build in Spring+Flow skeleton starter project with JS customisations.

PR #1707, opened 2023-12-12
test: add tests for waitForVaadin

tltv:
Adds a new hook in waitForVaadin while loop for tests. New tests verifies that waiting stops properly after client flag `devServerIsNotLoaded` is changed or client's `isActive()` return value changes.

Related to: #1705

### DIFF
--- a/vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
+++ b/vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
@@ -8,7 +8,10 @@
  */
 package com.vaadin.tests;
 
+import java.lang.reflect.Field;
+
 import org.junit.jupiter.api.Assertions;
+import org.openqa.selenium.JavascriptExecutor;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.testUI.PageObjectView;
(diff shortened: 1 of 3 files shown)


vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
```diff
@@ -38,6 +41,18 @@ public void waitForVaadin_activeConnector_waits() {
...
     }
 
+    @BrowserTest
+    public void waitForVaadin_activeConnector_waitsUtilReady() {
+        openTestURL();
+        assertDevServerIsNotLoaded();
+        getCommandExecutor().executeScript(
+                "window.Vaadin.Flow.clients[\"blocker\"] = {isActive: () => true};");
+        setWaitForVaadinLoopHook(500,
+                "window.Vaadin.Flow.clients[\"blocker\"] = {isActive: () => false};");
```
mshabarov, line 51: This might be done with executing the JS `"window.Vaadin.Flow.clients[\"blocker\"] = {isActive: () => false};"` as a callback to `setTimeout()` method, so the clients get non-active after some timeout threshold. Then you don't need a reflection and waitForVaadin hook.

tltv: `waitForVaadin` blocks both server and browser very effiently. `setTimeout` (and also I've tried `setInterval`, `requestAnimationFrame`) are only executed after while loop is done, which is too late.

mshabarov: Got it, thanks.

mshabarov (approved):

PR #1713, opened 2023-12-14
upgrade to selenium 4.16.1

ZheSun88:
4.16.0 contains broken packages, so 4.16.1 is needed


mcollovati (approved):

ZheSun88:
ah.. seems 9.0 and 9.1 already has this change

PR #1718, opened 2024-01-05
update jetty.version to 10.0.13 (8.2)

ZheSun88:
match with flow 23.3


ZheSun88:
close on behalf of #1720

Issue #1655, opened 2023-07-18
Quarkus support in testbench


ErrorProne:
### What do I want

Working quarkus support for pro features like UiUnitTest

### Explain your problem

Two things, based on who reads this

**Testbench-Devs:**
There are multiple problems we found with quarkus support so far:
1. The route-lookup does not work
2. Quarkus `@TestSecurity` (https://quarkus.io/guides/security-testing) feature is not working (similiar to springs @WithMockUser)
3. I'm not 100% sure about that, but it looks like the BeanInstantiator also has to be wired by hand.

All of those should be solved with the workarounds in [1] but an official out-of-the-box support should be the goal.

The vaadin project and team does such an amazing work and the framework is wonderful to use. The same should hold true for testing.

We will be happy to provide feedback or input to any implementations for this.

**Project owner/Pro feature Support:**
A little rant: Before switching to Vaadin we talked to a few folks and very explicitly stated that we do test a lot and use quarkus. We do deploy applications into factory lines. These apps have to work stable, no exceptions. Naturally we were very interested in the pro Testbench features and have been told that quarkus is supported. Our experience right now is: that is not the case at all.
We got it working eventually but only by diving into the vaadin code and with help from Marco Collovati in this issue in the quarkus extension here [1].

So far it looks like we solved all the issues in [1] but for a pro feature with official quarkus support and no mention about these problems at all (e.g. https://vaadin.com/docs/latest/integrations/quarkus) is a very frustrating experience. Additionally at least I see testing in 2023 not as an nice-to-have addition.

[1] https://github.com/vaadin/quarkus/issues/121

tarekoraby:
Thank you @ErrorProne for the ticket. The TestBench team, I'm sure, will address the technical feasibility of the proposed solution.

But as the product manager, my sincere apologies for any false promises made regarding TestBench support with Quarkus. It is indeed reasonable to expect TestBench to work with Quarkus without the need to resort to any workarounds. And if, for some reason, this support turns out to be impossible or prohibitively expensive to implement, we should clearly emphasize that in the Quarkus and TestBench documentation pages. 

In all cases, we will aim to investigate possible ways of improving TestBench + Quarkus support.

tarekoraby:
Thanks for sharing the workarounds 🙏  That's very helpful. 

It's the summer holiday season here in Finland, but we will get back to you as soon as we can once the bulk of our developers return and had the opportunity to examine the issue.

ErrorProne:
[...]

PR #1728, opened 2024-01-25
fix: fallback to context classloader when loading class

mcollovati:
## Description

Currently classes are loaded with Class.forName, that uses the classloader from the calling class. However, this might not work on all environments because of different class loading setup (e.g in Quarkus testing). This change uses the current thread context class loader as fallback when Class.forName is not able to load a class.

Fixes vaadin/quarkus#139
Part of #1655

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


mcollovati:
Fix can be tests running a UI Unit test on a [Quarkus based application](https://github.com/vaadin/base-starter-flow-quarkus/pull/new/spikes/ui-unit-test).
Running `MainViewTest` should fail with a `ClassNotFoundException`, whereas it should pass when using the patched Testbench artifacts.

mshabarov (approved):

PR #1771, opened 2024-03-14
test: wait for dev bundle creation before starting tests

mcollovati:
Should prevent, or at least reduce flakyness, due to tests starting while dev-bundle creation is still in progress.

### DIFF
--- a/vaadin-testbench-integration-tests-junit5/pom.xml
+++ b/vaadin-testbench-integration-tests-junit5/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <project
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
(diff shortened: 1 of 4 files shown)


vaadin-testbench-integration-tests-junit5/pom.xml
```diff
@@ -184,6 +184,38 @@
...
+                        <configuration>
+                            <mainClass>com.vaadin.tests.WaitForDevBundle</mainClass>
+                            <classpathScope>test</classpathScope>
+                            <systemProperties>
+                                <projectProperties/>
+                                <!-- Tune dev bundle creation waiter parameter if needed -->
+                                <!--
+                                <systemProperty>
+                                    <key>test.waitForDevBundle.attempts</key>
+                                    <value>5</value>
```
tepi, line 207: nit: Should we have the same 'default' value in this example as the `WaitForDevBundle` class has? Same for latter one as well.

mcollovati: done

tepi (approved):

mcollovati:
@tepi thanks for fixing the other pom :+1:

PR #1784, opened 2024-04-12
feat: add conditional (non-attribute) selectors to ElementQueries

### LINKED ISSUE #1183: Add additional selectors
Writing a testbench test requires a solid understanding of how a given Vaadin application is structured. The DX could be improved by allowing more general selectors.
For example `click("foo")` could click any element having the text "foo" (this could be a radio buttons or regular buttons for instance) as a text/caption.
Another example would be a `near` selector that simply selects the first element of a given type following a specific selection. For instance `click(checkbox(near("search"))`.

Both would allow for a more natural selection of elements.
(Examples were taken from https://taiko.gauge.org/#smart-selectors)

joelpop:
## Description

In contrast to `ComponentQuery`s of unit testing TestBench, `ElementQuery`s in E2E testing TestBench have (up until now) only provided attribute-based selector methods. This was expanded upon in PR #1774 to provide selector methods for common attributes, such as those for class names, theme, and id. This PR adds condition-based selector methods to `ElementQuery` to mirror and expand upon those of `ComponentQuery`. Namely:

- `withCondition`
- `withPropertyValue[Containing]`
- `withLabel[Containing]`
- `withPlaceholder[Containing]`
- `withCaption[Containing]`
- `withText[Containing]`

Elements having their labels in their text (such as buttons) must implement a new, method-less interface `HasLabelAsText` in their element tester (such as `NativeButtonElement` for the `NativeButton` component and `ButtonElement` for the `Button` component) in order for the `withCaption[Containing]` selectors to work for them.

This PR brings the E2E testing selectors up to par with the unit testing selectors. The only one missing is `withValue`.

Fixes #1183 Add Additional Selectors

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -400,6 +406,354 @@ public ElementQuery<T> withoutTheme(String theme) {
...
+            }
+            if (element instanceof HasPlaceholder hasPlaceholder) {
+                var placeholder = hasPlaceholder.getPlaceholder();
+                if (!placeholder.isEmpty()) {
+                    return comparison.test(placeholder, text);
+                }
+            }
+            if (element instanceof HasLabelAsText labelAsText) {
+                var label = labelAsText.getText();
+                if (!label.isEmpty()) {
```
mcollovati, line 646: Looking at the Javadoc of `WebElement.getText()` there is no guarantee that the returned value is not null, so maybe better add a null check

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
mcollovati, line 1: Talking about the selector methods that take a `BiPredicate`, I wonder if it would be better to add a note on the Javadoc to make explicit what the two parameters represent.
It looks to me a bit confusing right now.
For example, looking at `withLabel`, the parameter Javadoc says `the comparison to use when comparing the text with the label`, but the actual implementation provides them to the predicate in the opposite order `comparison.test(label, text)`.

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabel("one")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabel("two")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
```
mcollovati, line 323: I would probably change the view so that search for `one` and `two` returns a different number of items. Having the same assertion may hide a bug.
Also, it could be good to add a search for "none", asserting it gets an empty collection.

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withPlaceholder("one")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withPlaceholder("two")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
```
mcollovati, line 355: Same as above, maybe better having different size for the two searches, plus adding a "no elements found" case

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabelContaining("n")
+                .all();
+        Assertions.assertEquals(3, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabelContaining("o")
+                .all();
+        Assertions.assertEquals(4, labelPlaceholderElements.size());
```
mcollovati, line 339: What about adding a "no elements found" case?

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withPlaceholderContaining("w")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withPlaceholderContaining("o")
+                .all();
+        Assertions.assertEquals(4, labelPlaceholderElements.size());
```
mcollovati, line 371: What about adding a "no elements found" case?

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+                .withLabel("one")
+                .withPlaceholder("two")
+                .all();
+        Assertions.assertEquals(1, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabel("two")
+                .withPlaceholder("one")
+                .all();
+        Assertions.assertEquals(1, labelPlaceholderElements.size());
```
mcollovati, line 389: Same as above, maybe better having different size for the two searches, plus adding a "no elements found" case

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+                .withLabelContaining("n")
+                .withPlaceholderContaining("w")
+                .all();
+        Assertions.assertEquals(1, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withLabelContaining("o")
+                .withPlaceholderContaining("o")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
```
mcollovati, line 407: What about adding a "no elements found" case?

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withCaption("one")
+                .all();
+        Assertions.assertEquals(3, labelPlaceholderElements.size());
+
+        labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withCaption("two")
+                .all();
+        Assertions.assertEquals(3, labelPlaceholderElements.size());
```
mcollovati, line 423: Same as above, maybe better having different size for the two searches, plus adding a "no elements found" case

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+        List<LabelPlaceholderElement> labelPlaceholderElements = view.$(LabelPlaceholderElement.class)
+                .withCaptionContaining("Special")
+                .all();
+        Assertions.assertEquals(2, labelPlaceholderElements.size());
+
+        List<NativeButtonElement> nativeButtonElements = view.$(NativeButtonElement.class)
+                .withCaptionContaining("Special")
+                .all();
+        Assertions.assertEquals(2, nativeButtonElements.size());
```
mcollovati, line 444: Same as above, maybe better having different size for the two searches, plus adding a "no elements found" case

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+    @BrowserTest
+    public void textMatches() {
+        openTestURL();
+        TemplateViewElement view = $(TemplateViewElement.class).waitForFirst();
+
+        List<NativeButtonElement> nativeButtonElements = view.$(NativeButtonElement.class)
+                .withText("Button with special id")
+                .all();
+        Assertions.assertEquals(1, nativeButtonElements.size());
```
mcollovati, line 455: What about adding a "no elements found" case?

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/ElementQueryIT.java
```diff
@@ -305,4 +307,163 @@ void allWithoutLightTheme() {
...
+
+    @BrowserTest
+    public void textContains() {
+        openTestURL();
+        TemplateViewElement view = $(TemplateViewElement.class).waitForFirst();
+
+        List<NativeButtonElement> nativeButtonElements = view.$(NativeButtonElement.class)
+                .withTextContaining("Special")
+                .all();
+        Assertions.assertEquals(2, nativeButtonElements.size());
```
mcollovati, line 466: What about adding a "no elements found" case?

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
mcollovati: Another general question for the new selectors that take the expected string value: how should we consider a blank or null input? Is it an illegal argument?

mcollovati (changes requested):

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: I considered that during implementation and came to the conclusion that we should allow both blank and null values. We should not restrict the developers from making those selections or comparisons.

joelpop: Now that I had another chance to look back through the comparison implementations, realize that null values never make sense in comparisons, but I don't think we should prevent them or give them any special handling if the developer supplies a null value—it just won't compare positively. I have cleaned up some code as a result. For `getText()`, since we can't be guaranteed the return value is not null, I have wrapped them as `Objects.requireNonNullElse(element.getText(), "")`.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -400,6 +405,377 @@ public ElementQuery<T> withoutTheme(String theme) {
...
+     * @param comparison
+     *          the comparison to use when comparing the caption
+     *          against the supplied text
+     *          (i.e., {@code comparison.test(elementCaption, text)}
+     * @return this element query instance for chaining
+     *
+     * @see #withCaption(String)
+     * @see #withCaptionContaining(String)
+     */
+    public ElementQuery<T> withCaption(String text, BiPredicate<String, String> comparison) {
```
mcollovati, line 651: It looks like to me that this method behaves differently from others regarding handling of empty string for `text` parameter.
Here, if the element has an empty label/caption/text, the element is rejected without applying the `comparison`, whereas in other methods there's no such restriction.
For example, in `withPlaceholder` the `comparison` is applied without first checking if the element placeholder is empty.

We should always apply comparison, on the first non-empty value from the element, or empty if no attribute is valued.
Otherwise, we should not allow empty string for the `text` parameter, throwing an exception.
Probably the first one is more consistent with the rest of the other selectors.

joelpop: This difference in behavior is intentional. For `withLabel`, `withPlaceholder`, and `withText', we are explicitly looking at one and only one of those three places. However, `withCaption` is a convenience method—we don't know in which one of those three places the caption will be found, which is why when we find that particular value in the element to be empty, we move on to look at the next one.

I guess the one thing that makes less sense is that even though you should be able to match on an empty text value for `withLabel`, `withPlaceholder`, and `withText', what should that mean for captions? The way it works now is that if you provide an empty text value, it will incorrectly select an element with an empty label even if the element has a placeholder.

I think that is the thing I need to revise—in the special case when the input text value is empty, then all three places should be empty. I will make this adjustment. In addition, since none of the places can now ever have a null text value, I will disallow null text for each of them.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -400,6 +408,395 @@ public ElementQuery<T> withoutTheme(String theme) {
...
+        return withPlaceholder(text, String::contains);
+    }
+
+    /**
+     * Requires the element's caption (i.e., label, placeholder, or text label)
+     * to satisfy the given comparison with the supplied text.
+     * <p>
+     * If the element has a label, the label value is used.
+     * If the element can have a label, but its value is empty and it has a placeholder,
+     * the placeholder value is used.
```
mcollovati, line 630: This statement is a bit confusing to me: it looks like that `placeholder` is used only if the element has an empty label, not if it does not support a label.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -400,6 +408,395 @@ public ElementQuery<T> withoutTheme(String theme) {
...
+        return withCondition(element -> {
+            // special case when text is empty,
+            // so if they both exist, both label and placeholder must be empty
+            if (text.isEmpty() &&
+                    element instanceof HasLabel hasLabel &&
+                    element instanceof HasPlaceholder hasPlaceholder) {
+                var label = hasLabel.getLabel();
+                var placeholder = hasPlaceholder.getPlaceholder();
+                return label.isEmpty() && placeholder.isEmpty();
+            }
```
mcollovati, line 670: This special case must be documented, since it might not be obvious to a developer.

mcollovati (changes requested):

mcollovati (approved):

PR #1786, opened 2024-04-16
feat: rework "containing" selectors via addition of remaining attribute operators

### LINKED ISSUE #1554: ElementQuery - attributeContains rework
I've been caught out a couple of times by ElementQuery#attributeContains by the fact that it's not like the String 'contains' method.
"attributeContains" makes me think that whatever I provide exists anywhere within the actual attribute value as a substring (aka, `*=` matcher rather than the current `~=` matcher).

I would suggest this method be split.
`~=` could be a "attributeContainsEntry" or "attributeHasPart" method, while
`*=` would be used in "attributeContains".

It's rough that there's no way to implement our own extensions or forks of ElementQuery in order to add this functionality ourselves, but perhaps that's a submission for another time.

joelpop:
## Description

Refactored name of "xContaining" attribute selector methods to "xContainingToken".

Added true "xContaining" attribute selector methods that match on substrings instead of on tokens.

Added remaining attribute matching operators and a corresponding selector method to utilize them. Refactored existing attribute-based selector methods to use this method.

Fixes #1554 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
+++ b/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
@@ -1,5 +1,5 @@
-/**
- * Copyright (C) 2000-2022 Vaadin Ltd
+/*
+ * Copyright (C) 2000-2024 Vaadin Ltd
  *
  * This program is available under Vaadin Commercial License and Service Terms.
  *
@@ -27,6 +27,15 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.CONTAINS;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.CONTAINS_TOKEN;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.EXISTS;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.MATCHES_EXACTLY;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.NOT_CONTAINS;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.NOT_CONTAINS_TOKEN;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.NOT_EXISTS;
+import static com.vaadin.testbench.ElementQuery.AttributeMatch.Operator.NOT_MATCHES_EXACTLY;
+
 /**
  * Query class used for finding a given element inside a given search context.
  * <p>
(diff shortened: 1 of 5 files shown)


vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -48,33 +57,153 @@ public class ElementQuery<T extends TestBenchElement> {
...
+             * Attribute value begins with a space/hyphen-separated prefix
+             * that matches the given value.
+             * The prefix must be either the entire attribute value or
+             * the leading hyphen-separated segments of the attribute value.
+             */
+            CONTAINS_PREFIX("|="),
+            /**
+             * Attribute value begins with the given value.
+             */
+            BEGINS_WITH("^="),
```
mcollovati, line 89: What about `STARTS_WITH`? It might be more familiar to Java developers.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -48,33 +57,153 @@ public class ElementQuery<T extends TestBenchElement> {
...
+            MATCHES_EXACTLY("="),
+            /**
+             * Attribute value contains the given value.
+             */
+            CONTAINS("*="),
+            /**
+             * Attribute value contains a space-separated token
+             * that matches the given value.
+             */
+            CONTAINS_TOKEN("~="),
```
mcollovati, line 78: nit: Just a personal preference: what about `CONTAINS_ITEM` or `CONTAINS_WORD`?

For example, I've found this description for the selector, using `word` term

> Represents elements with an attribute name of attr whose value is a whitespace-separated list of words, one of which is exactly value.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -48,33 +57,153 @@ public class ElementQuery<T extends TestBenchElement> {
...
+            Operator(String op, boolean neg) {
+                this.op = op;
+                this.neg = neg;
+            }
+
+            Operator(String op) {
+                this(op, false);
+            }
+
+            public String getOp() {
```
mcollovati, line 138: As this becomes a public API, we should have a more friendly name, e.g. `getOperator()` :grin: 
Also needs Javadoc.
The same goes for `isNeg()`

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -48,33 +57,153 @@ public class ElementQuery<T extends TestBenchElement> {
...
+         *
+         * @deprecated use {@link #AttributeMatch(String, Operator, String)}
+         */
+        @Deprecated(since = "9.3")
+        public AttributeMatch(String name, String operator, String value) {
+            this(name,
+                    Arrays.stream(Operator.values())
+                            .filter(op -> op.getOp().equals(operator))
+                            .findFirst()
+                            .orElse(EXISTS),
```
mcollovati, line 183: Previously, we (probably mistakenly) accepted any operator the developer provided; now we turn unknown operators into `EXISTS`.

We should either document the new behavior or throw an exception.

In both cases, this is a breaking change.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -258,41 +438,44 @@ public ElementQuery<T> withAttribute(String attribute, String value) {
...
      *            the token to look for
      * @return this element query instance for chaining
      *
      * @see #withAttribute(String, String)
+     * @see #withAttributeContaining(String, String)
      */
-    public ElementQuery<T> withAttributeContaining(String attribute, String value) {
-        attributes.add(new AttributeMatch(attribute, "~=", value));
-        return this;
+    public ElementQuery<T> withAttributeContainingToken(String attribute, String token) {
```
mcollovati, line 453: The example in the Javadoc should probably mention `withAttributeContainingToken` instead of `withAttributeContaining`

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -301,10 +484,34 @@ public ElementQuery<T> withoutAttribute(String attribute) {
...
+     * @see #withoutAttributeContainingToken(String, String)
      */
     public ElementQuery<T> withoutAttribute(String attribute, String value) {
-        attributes.add(new AttributeMatch(attribute, "!=", value));
-        return this;
+        return withAttribute(attribute, value, NOT_MATCHES_EXACTLY);
+    }
+
+    /**
+     * Selects on elements not having the given attribute containing the given token.
```
mcollovati, line 494: `token` could be misleading here
```suggestion
     * Selects on elements not having the given attribute containing the given text.
```

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -316,29 +523,30 @@ public ElementQuery<T> withoutAttribute(String attribute, String value) {
...
      *            the token to look for
      * @return this element query instance for chaining
      *
      * @see #withoutAttribute(String, String)
+     * @see #withoutAttributeContaining(String, String)
      */
-    public ElementQuery<T> withoutAttributeContaining(String attribute, String value) {
-        attributes.add(new AttributeMatch(attribute, "!~=", value));
-        return this;
+    public ElementQuery<T> withoutAttributeContainingToken(String attribute, String token) {
```
mcollovati, line 538: Example code in Javadoc needs to be fixed

mcollovati (changes requested):

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: IMO, that was an unfortunate name choice if you want to maintain parallelism.
start → stop  ✅
begin → end  ✅
start → end  ❓

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Sure.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
mcollovati: I see. 
I just made a comparison with what we have in String class.
I'm fine also with begins. Feel free to choose what do you prefer

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Agreed. So there would be no confusion by name reuse, I changed the enum's name to Comparison and spelled out the member/getter names.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: I prefer the exception option, as invalid operators fail upon use, so should not be allowed, and we don't want them failing silently.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Good catch!

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Good catch!

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Good catch!

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
joelpop: Let's keep it the way it is.

vaadin-testbench-shared/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -55,34 +64,200 @@ public class ElementQuery<T extends TestBenchElement> {
...
+         *
+         * @deprecated use {@link #AttributeMatch(String, Comparison, String)}
+         */
+        @Deprecated(forRemoval = true, since = "9.3")
+        public AttributeMatch(String name, String operator, String value) {
+            this(name,
+                    Arrays.stream(Comparison.values())
+                            .filter(comp -> comp.getOperator().equals(operator))
+                            .findFirst()
+                            .orElseThrow(),
```
mcollovati, line 237: Better provide a detailed message here, suggesting using the other constructor

mcollovati (changes requested):

mcollovati (approved):

PR #1815, opened 2024-08-14
fix: Disable Chrome's search engine selection dialogue in LocalDriver

Ansku:
The search engine selection dialogue doesn't affect testing, or show up in screenshots, but it makes local debugging difficult by hiding content visually.


thevaadinman (approved):

Artur-:
Why was this done and merged for 5.3 when the workflow is always to merge into main and backport when needed?

Ansku:
Because different teams maintain the V8/7 branches and the Flow branches, and because this issue was blocking my attempts to test another issue in another project.

Issue #1817, opened 2024-08-23
Local Chrome driver in non-headless mode shows search engine selection


sissbruecker:
Since Chrome 127, the browser shows a search engine selection screen every time when running tests locally:
<img width="1189" alt="Bildschirmfoto 2024-08-23 um 09 16 42" src="https://github.com/user-attachments/assets/e105958f-3401-4183-9593-1fc33fb51448">

This is problematic because:
- It is bothersome if you want to debug some test in the browser
- In at least one case with Chart ITs the dialog seems to cause tests to fail

Testbench should automatically disable the search engine selection screen by setting the argument `--disable-search-engine-choice-screen` ([Reference](https://www.selenium.dev/blog/2024/chrome-browser-woes/)).

PR #1818, opened 2024-08-23
fix: disable chrome search engine selection

sissbruecker:
Fixes https://github.com/vaadin/testbench/issues/1817

### DIFF
--- a/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java
+++ b/vaadin-testbench-shared/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java
@@ -71,12 +71,14 @@ public class LocalDriver {
             }
             driver = new FirefoxDriver(options);
         } else if (BrowserUtil.isChrome(desiredCapabilities)) {
+            ChromeOptions options = new ChromeOptions();
             // Tells chrome not to show warning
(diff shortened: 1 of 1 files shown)


vaadin-testbench-shared/src/main/java/com/vaadin/testbench/parallel/setup/LocalDriver.java
```diff
@@ -71,12 +71,14 @@ static public WebDriver createDriver(
...
         } else if (BrowserUtil.isChrome(desiredCapabilities)) {
+            ChromeOptions options = new ChromeOptions();
             // Tells chrome not to show warning
             // "You are using an unsupported command-line flag:
             // --ignore-certifcate-errors".
             // #14319
-            ChromeOptions options = new ChromeOptions();
             options.addArguments("--test-type ");
+            // Disable search engine choice screen
+            options.addArguments("--disable-search-engine-choice-screen");
```
sissbruecker, line 81: I'm not sure if the search engine selection should be disabled for other types of tests too. 

However for local tests it's problematic because it seems that you can't customize the Chrome options in your test code - the method here ignores just ignores the capabilities passed in, so even if you would try to add this option in your test setup it doesn't work.

Artur- (approved):

PR #1821, opened 2024-09-06
chore: DevelopmentServerLaunder & WebDriverManager for integration tests

Ansku:
Also update the IT tests to use new deployment paths, convert to using
direct imports to Assert methods, and switch to catching
ElementNotInteractableException for added compatibility
(ElementNotVisibleException extends it).


thevaadinman (approved):
I don't detect anything inherently dangerous in here, though the dependency list could probably do with some pruning. Regardless, approved.

PR #1824, opened 2024-09-12
chore: Javadoc fixes

Ansku:


vaadin-testbench-core/src/main/java/com/vaadin/testbench/ElementQuery.java
```diff
@@ -279,6 +295,8 @@ public ElementQuery<T> context(SearchContext searchContext) {
     /**
      * Return the {@link SearchContext} previously set with the {@link #in}
      * call.
+     *
+     * @return context
```
thevaadinman, line 299: Doesn't really add much information

vaadin-testbench-core/src/main/java/com/vaadin/testbench/HasTestBenchCommandExecutor.java
```diff
@@ -23,8 +23,8 @@ public interface HasTestBenchCommandExecutor extends HasSearchContext {
     /**
      * Return a reference to the related {@link TestBenchCommandExecutor}
      * instance.
-     * 
-     * @return
+     *
+     * @return executor
```
thevaadinman, line 27: Doesn't add much information

vaadin-testbench-core/src/main/java/com/vaadin/testbench/Parameters.java
```diff
@@ -158,6 +160,7 @@ public static boolean isScreenshotComparisonCursorDetection() {
      * Sets the directory to search for reference images.
      *
      * @param screenshotReferenceDirectory
+     *            directory path
```
thevaadinman, line 163: This adds SOME information, but perhaps more would be better?

vaadin-testbench-core/src/main/java/com/vaadin/testbench/Parameters.java
```diff
@@ -175,6 +178,7 @@ public static String getScreenshotReferenceDirectory() {
      * Sets the directory where error screen shots are stored.
      *
      * @param screenshotErrorDirectory
+     *            directory path
```
thevaadinman, line 181: See above

vaadin-testbench-core/src/main/java/com/vaadin/testbench/Parameters.java
```diff
@@ -223,8 +230,10 @@ public static void setCaptureScreenshotOnFailure(
     }
 
     /**
-     * @return whether to capture a screen shot when a test fails or not.
+     * @return whether to capture a screen shot when a test fails or not
```
thevaadinman, line 233: This should probably have a slightly lengthier description, as above

vaadin-testbench-core/src/main/java/com/vaadin/testbench/TestBench.java
```diff
@@ -119,6 +119,8 @@ public static TestBenchElement createElement(WebElement webElement,
      * TestBenchCommandExecutor. This feature is advanced and potentially
      * dangerous.
      *
+     * @param <T>
+     *            the type of the class
```
thevaadinman, line 123: As a description, this is confusing. "Type of the class"? Should it perhaps inherit from something or have special traits?

vaadin-testbench-core/src/main/java/com/vaadin/testbench/screenshot/ImageUtil.java
```diff
@@ -70,7 +70,8 @@ public static String encodeImageToBase64(BufferedImage image) {
      * Get luminance value for the given rgb value.
      * 
      * @param rgb
-     * @return
+     *            rgb value
```
thevaadinman, line 73: I get the feeling that we might want to explain that the red, green and blue byte values should be packed into an integer at given offsets somewhere here..

thevaadinman (commented):
Some comments, but in general this seems fine

vaadin-testbench-core/src/main/java/com/vaadin/testbench/screenshot/ImageUtil.java
thevaadinman: something like `The RGB value is assumed to be taking up the low 24 bits of the integer sample, with red shifted by 16 bits, green by 8 bits and blue by 0 bits, ORed together. This function will not return the correct values for other channel orders.`

thevaadinman (approved):

Ansku:
Added some tweaks, left out the 'no added value' changes because the main purpose is to get rid of the Javadoc errors in build logs and the method description already contains the extended information.

thevaadinman (approved):

PR #1845, opened 2025-01-13
fix: fix date/time/datetime picker tester validation

### LINKED ISSUE #1840: DatePickerTester and DateTimePickerTester try to find isInvalid
The `DatePickerTester.setValue` method tries to find a method `isInvalid` with a parameter `LocalDate` which does not exists. Same on `DateTimePickerTester.setValue` but with a `LocalDateTime`. Seems like they should use the isInvalid without parameter.

mcollovati:
Validation API of Date/Time/DateTime picker has been changed since Vaadin 24.5, and the UI Unit testers are failing with recent Vaadin version. This change makes the testers compatible with all Vaadin 24 series.

Fixes #1840


mcollovati:
Tested the change on 24.5, 24.6 and 24.7 and validation passed (#1846).

tepi (approved):

PR #1849, opened 2025-01-16
fix: handle type not found in pretty print

mcollovati:
Vaadin 24.7 introduces helper methods for Spring Data integration. This caused an issue with PrettyPrinter introspection when UI Unit test is used in non Spring projects.
This change adds a safe fallback for members introspection if kotlin inspector fails.


mshabarov (approved):
Looks good to me, however, I suggest to add two minor things:
1. A comment that explains the reason for `TypeNotPresentException`
2. give meaningful var names to `f` and `m`.

mshabarov (approved):

Issue #1843, opened 2025-01-13
WebElement getAttribute is deprecated, TBElement should mark it as deprecated, too and override replacement methods


stefanuebe:
As described in the title, the Selenium WebElement has deprecated its `getAttribute()` method and recommends to use `getDomAttribute/Property()` instead. The TestBenchElement variant should also be deprecated and have a respective comment. 

Additionally, it needs to override `getDomAttribute/Property` to not throw an `UnsupportedOpException`

PR #1865, opened 2025-02-04
chore: deprecate getAttribute

tepi:
Fixes https://github.com/vaadin/testbench/issues/1843


mcollovati (approved):
Looks good, and it seems not to break anything, at least in Flow tests.

PR #1872, opened 2025-02-18
fix: allow comma in broserOptions

manolo:
There are options that needs a comma like in `--window-size=nnn,nnn`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <Implementation-Version>${project.version}</Implementation-Version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <selenium.version>4.28.0</selenium.version>
+        <selenium.version>4.29.0</selenium.version>
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
         <spring.boot.version>3.4.2</spring.boot.version>


mcollovati (commented):
It looks like a potential breaking change, if we don't allow comma anymore as a separator.
For example, if users' CI jobs are using comma separated values, their build might start failing, and it could be difficult to find the issue.
On the other hand, the fix is straightforward, but it must be highlighted in the release notes.

Another option is to improve the regular expressions to consider only spaces and commas not surrounded by double quotes (e.g. `-foo, -bar,--window-size="100,100"`), but the regex would become quite complicated, and probably it is not worth it.

If we agree on disallowing commas as separator, the Javadoc must be updated as well.

mcollovati:
We could also take the opportunity to add a test in `vaadin-testbench-core/src/test/java/com/vaadin/testbench/ParametersTest.java`, something like the following (adapted to the rules we want to implement)

```java
    @Test
    public void setChromeOptionsSystemProperty() {
        try {
            Parameters.setChromeOptions("--foo");
            Assert.assertArrayEquals(new String[] { "--foo", },
                    Parameters.getChromeOptions());
            Parameters.setChromeOptions("--foo --bar -x");
            Assert.assertArrayEquals(new String[] { "--foo", "--bar", "-x" },
                    Parameters.getChromeOptions());
            Parameters.setChromeOptions("--foo,--bar,-x");
            Assert.assertArrayEquals(new String[] { "--foo", "--bar", "-x" },
                    Parameters.getChromeOptions());
            Parameters.setChromeOptions("--window-size=\"nnn,nnn\"");
            Assert.assertArrayEquals(new String[] { "--window-size=nnn,nnn", },
                    Parameters.getChromeOptions());
            Parameters.setChromeOptions("--window-size=\"nnn,nnn\" --bar,-x");
            Assert.assertArrayEquals(
                    new String[] { "--window-size=nnn,nnn", "--bar", "-x" },
                    Parameters.getChromeOptions());
            Parameters.setChromeOptions("");
            Assert.assertNull(Parameters.getChromeOptions());
        } finally {
            Parameters.setChromeOptions(null);
        }
    }
```

mcollovati (approved):

PR #1906, opened 2025-04-03
fix: Focus on click

TatuLund:
Mere JS call to click does not focus the element. Real life action of clicking element always focuses it. This is important if in the test story it is important that clicking Button e.g. blurs TextField, which in turn triggers value change in TextField.


mcollovati (commented):
At a first look, the change seems fine, but I'd like to do some quick tests.
What I do not like who the `getChildrenTest` is modified. We are renaming it and adding multiple assertion for different use cases, increasing the maintenance complexity. I would prefer a separated test for the focus handling.

TatuLund:
> What I do not like is how the `getChildrenTest` is modified.

That is about the other PR, but yes I fixed that there.

mcollovati:
> That is about the other PR, but yes I fixed that there.

Great. Let's merge that one and the update this according to the changes

TatuLund:
> I still do not like this clickingButtonWillFocusIt that mixes tests for focus and children and would prefer separated tests.

I was confused what you mean, but now fixed. It was remains of merge conflict.

PR #1914, opened 2025-04-08
chore: Rename all TB3 things in test classes to TB5 things

Ansku:


thevaadinman (approved):
Interesting change, but worth propagating to be explicit about not relying on old stuff in Framework, at least

PR #1920, opened 2025-04-16
Cherry pick from 9.4 to 8.3

mshabarov:
DO NOT SQUASH


mshabarov:
@TatuLund do you think all of these commits make sense to backport ?

TatuLund:
I would think so.

mcollovati (approved):
Since commits will not be squashed, comments can be addressed in another PR

PR #1958, opened 2025-06-27
chore: Stop using default versions in IT tests

Ansku:
Use Linux as the default Platform.

Also update reference screenshots and their names to match the current
CI environment, and remove outdated PhantomJS screenshots.


thevaadinman (approved):
Changes look reasonable

PR #2028, opened 2025-10-03
chore: Upgrade to JUnit 6

mshabarov:


Artur- (commented):
How can the junit 5 version be 6? What happens with all existing projects that use 5?

mshabarov:
To be designed how we recommend using TestBench with JUnit 6, but this PR can be closed I think.

PR #2036, opened 2025-10-07
fix: Make ComboBoxTester.selectItem respect native filtering (#2003)

### LINKED ISSUE #2003: ComboBoxTester.selectItem doesn't respect ComboBox native filtering.
We have a ComboBox that performs custom filtering

```
        setItems(
            { item, filterText ->
                ... custom filtering logic removed ...
            },
            myRepository.findAll().toList()
        )
```

Calling ComboBoxTester.selectItem doesn't respect this filtering and does filtering of its own

```
        final List<Y> filtered = suggestionItems.stream().filter(
                item -> selection.equals(itemLabelGenerator.apply(item)))
                .collect(Collectors.toList());
```
When calling the selectItem API it should respect the custom filtering behavior of the ComboBox.

Artur-:
The selectItem method now applies the filter text before searching for items, ensuring it respects the ComboBox's custom filtering logic set via setItems(ItemFilter, Collection). This matches the behavior a user would experience when typing in the ComboBox.

- Applied filter before selection in ComboBoxTester.selectItem
- Applied same fix to MultiSelectComboBoxTester.selectItem
- Clear filter after selection to restore original state
- Added tests to verify the fix works with custom filtering

Fixes #2003

### DIFF
--- a/vaadin-testbench-unit-junit5/src/test/java/com/vaadin/flow/component/combobox/ComboBoxCustomFilteringTesterTest.java
+++ b/vaadin-testbench-unit-junit5/src/test/java/com/vaadin/flow/component/combobox/ComboBoxCustomFilteringTesterTest.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright (C) 2000-2025 Vaadin Ltd
+ *
+ * This program is available under Vaadin Commercial License and Service Terms.
+ *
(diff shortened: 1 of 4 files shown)


mcollovati:
Review is pending on https://github.com/vaadin/testbench/issues/2003#issuecomment-3381249896.
It could be that the change is not necessary.

PR #2042, opened 2025-10-13
fix: add Signals support for unit tests with controllable task flushing

### LINKED ISSUE #1968: Make sure UI unit tests using signals work out of the box.
We should make sure everything related to signals work out of the box for UI unit tests. We should make sure we have test coverage for such test cases.

For reference, two things are needed for signals to function:
* Initializing the `SignalEnvironment`. This is normally done through  `VaadinService::init`.
* `ComponentEffect` only works for components that are attached.

mcollovati:
Introduce a test-only TestSignalEnvironment that captures all Signals-related tasks when no VaadinSession/VaadinService is available (e.g., background threads).
Tests can deterministically flush these tasks via new runPendingSignalsTasks helpers in BaseUIUnitTest.

Fixes #1968


tltv (approved):
LGTM

mcollovati:
@mshabarov I updated a little bit the test

PR #2059, opened 2025-11-06
refactor: remove gwt-elemental dependency

sissbruecker:
## Description

Removes `gwt-elemental` dependencies and replaces them with Jackson where necessary.

The dependency is still pulled in by `com.vaadin:license-checker` so that could use changes as well.

## Type of change

- Refactor


Artur-:
I have a hard time seeing when it would be good to use the wrong type in a test... like let's assert that `button.disableOnClick` equals `1`

PR #2072, opened 2025-11-11
ci: restructure CI workflow to optimize SauceLabs usage and build times

Artur-:
Completely restructure the validation workflow to separate concerns and
respect SauceLabs concurrent session limits while maximizing parallelism.

## New Job Structure

1. **format-check** - Code style validation (unchanged)
2. **build** - Build entire project once, cache artifacts for downstream jobs
3. **unit-tests** - Run all unit tests using cached build (no SauceLabs)
4. **integration-tests** - Run SauceLabs browser tests using cached build
   - Matrix: JUnit 4, 5, 6 variants
   - Sequential execution within workflow (max-parallel: 1)
   - Global concurrency control across all PRs

## Key Improvements

**Build once, use everywhere:**
- Build job runs `mvn clean install -DskipTests` once
- Artifacts cached with run-specific key: `github.run_id`
- Both unit-tests and integration-tests restore from cache
- Eliminates redundant builds (was building 3x, once per JUnit version)

**Independent unit test execution:**
- Runs `mvn test` separately from integration tests
- No SauceLabs dependency (fast, cheap, can parallelize)
- Separate test reports: surefire-reports (unit tests)
- Added TestBench license setup for Pro feature validation

**Optimized integration tests:**
- Runs `mvn verify -DskipTests -pl <module> -am -P validation`
- `-DskipTests` skips unit tests (already ran in separate job)
- Only runs failsafe integration tests (*IT.java)
- Uses full SauceLabs parallelism: `testsInParallel=5`, `forkCount=5`
- Separate test reports: failsafe-reports (integration tests)

**SauceLabs concurrency control:**
- Job-level concurrency group: `saucelabs-testbench`
- Ensures only one set of integration tests runs globally
- Matrix strategy runs JUnit 4/5/6 sequentially (max-parallel: 1)
- Total concurrent sessions: exactly 5 (within SauceLabs limit)

## Resource Utilization

**Multiple PRs can run in parallel:**
- format-check: ✅ parallel across all PRs
- build: ✅ parallel across all PRs
- unit-tests: ✅ parallel across all PRs
- integration-tests: ❌ queued globally (SauceLabs constraint)

**Performance benefits:**
- Faster feedback: format/build/unit-tests run immediately in parallel
- Faster builds: build once vs 3x (33% of previous time)
- Better SauceLabs utilization: full 5-session parallelism per test run
- No wasted concurrent sessions from multiple workflows

## Maven Cache Strategy
Cache key includes `github.run_id` for per-workflow isolation:
- build job: creates cache with `actions/cache@v4`
- unit-tests job: restores cache with `actions/cache/restore@v4`
- integration-tests job: restores cache with `actions/cache/restore@v4`

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -235,15 +235,15 @@ jobs:
             **/target/failsafe-reports/TEST-*.xml
           retention-days: 7
 
-  test-results:
-    name: Test Results
+  validation-status:
+    name: Validation Status
     permissions:
       actions: write
       issues: read
       checks: write
       pull-requests: write
     if: always()
-    needs: [unit-tests, integration-tests]
+    needs: [format-check, build, unit-tests, integration-tests]
     runs-on: ubuntu-latest
 
     steps:
@@ -266,31 +266,17 @@ jobs:
           check_name: Test Results
           comment_mode: failures
 
-      - name: Check test job results
-        run: |
-          if [ "${{ needs.unit-tests.result }}" != "success" ] || \
-             [ "${{ needs.integration-tests.result }}" != "success" ]; then
-            echo "🚫 One or more test jobs failed"
-            exit 1
-          fi
-          echo "All tests completed successfully"
-
-  validation-status:
-    name: Validation Status
-    needs: [format-check, build, test-results]
-    runs-on: ubuntu-latest
-    if: always()
-
-    steps:
       - name: Check all jobs status
         run: |
           echo "format-check: ${{ needs.format-check.result }}"
           echo "build: ${{ needs.build.result }}"
-          echo "test-results: ${{ needs.test-results.result }}"
+          echo "unit-tests: ${{ needs.unit-tests.result }}"
+          echo "integration-tests: ${{ needs.integration-tests.result }}"
 
           if [ "${{ needs.format-check.result }}" != "success" ] || \
              [ "${{ needs.build.result }}" != "success" ] || \
-             [ "${{ needs.test-results.result }}" != "success" ]; then
+             [ "${{ needs.unit-tests.result }}" != "success" ] || \
+             [ "${{ needs.integration-tests.result }}" != "success" ]; then
             echo "One or more validation jobs failed"
             exit 1
           fi


.github/workflows/validation.yml
```diff
@@ -91,19 +195,16 @@ jobs:
...
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           SAUCE_TUNNEL_ID: ${{ github.run_id }}-${{ github.run_number }}
         run: |
-          mvn javadoc:javadoc verify \
+          mvn verify \
+            -pl ${{ matrix.module }} -am \
             -P validation \
-            -Dtestbench.javadocs \
+            -Dsurefire.skip=true \
```
mcollovati: I can't find this property documented anywhere, nor used by the project POMs.

mcollovati:
It looks like that the workflow is now missing a final step that depends on all other jobs to be used as "required status check" in the protected branches configuration.
Currently, a "Build and Test" check is expected, but it does not exist anymore in the workflow.

Side question: the workflow acts on `pull_request` event, but for example, in Flow we use `pull_request_target`. Is this intended? Ping also @ZheSun88 that may have some insights here.

Artur-:
We should use pull_request_target in the end. Should we change everything in this one PR or do it separately?

Artur-:
Changed to pull_request_target

Artur-:
Let's see if this approach works - not using a profile

mcollovati:
Looks good now. Hopefully, one day we can parallelize the integration tests.

Just a last note: after the change gets merged and backported to all branches, we have to update the protected branches settings to wait for `validation-status` check instead of the current one.

mcollovati (approved):

mcollovati:
There's an issue in the "Merge test results" step

```
Found 4 artifact(s)
Error: No artifacts found matching pattern 'unit-test-results
integration-test-results-*'
```

Artur-:
Yes, let's fix that and then maybe remove the validation status step and instead use the "test result" step

mcollovati:
@ZheSun88 which are the maintained branches of this repo?

PR #2093, opened 2025-11-24
chore: change license-checker to provided scope

tltv:
Changes license-checker compile scope to provided to ensure that production package doesn't include transitive `license-checker` dependency via `test` scoped testbench dependency, when it's also added transitively via `optional` `vaadin-dev` dependency.

This would be breaking change for those setups that uses testbench for production package but don't have license-checker dependency directly or transitive in other manners. Fix is to add license-checker directly or via `optional` `vaadin-dev`.


tltv:
In starter apps it would be `vaadin-dev` providing license-checker. 

Instead of `provided`, it could probably be also changed to `optional`.

tltv:
This is most likely not needed.

PR #2095, opened 2025-11-26
chore: Use JDK for base64

mstahv:


mstahv:
Created as draft as commons-codec is good indicator while investigating the infamous "TB dependency weirdness"...

mcollovati (approved):

PR #2120, opened 2026-01-23
fix: Fix click count in TestBenchElement#click()

TatuLund:


mcollovati:
To avoid potential incompatibilities, I would not change the `scrollIntoView` method. What about adding a new one that takes the options Map as a parameter?

PR #2159, opened 2026-03-02
fix: use Selenium Actions for click()

mcollovati:
TestBenchElement.click() was using JavaScript's element.click(), which does not produce the same browser events as a real user click. This made click() behave differently from click(x, y), doubleClick(), and contextClick(), which all use Selenium Actions and correctly simulate user interactions. Side effects include missing or incorrect browser event properties such as click count.

Replace the default click() implementation with Selenium Actions.click(), consistent with the other click methods. The old JavaScript-based click logic is extracted into a new javascriptClick() method for cases where Selenium cannot reach the element directly.

A useJavascriptClick parameter is added to Parameters for backwards compatibility, allowing existing projects to restore the old behavior via system property or programmatic setter.

Fixes #2155

### DIFF
--- a/vaadin-testbench-core-junit5/src/test/java/com/vaadin/testbench/ParametersTest.java
+++ b/vaadin-testbench-core-junit5/src/test/java/com/vaadin/testbench/ParametersTest.java
@@ -30,6 +30,7 @@ public class ParametersTest {
         Assertions.assertEquals(false,
                 Parameters.isScreenshotComparisonCursorDetection());
         Assertions.assertFalse(Parameters.isHeadless());
+        Assertions.assertFalse(Parameters.isUseJavascriptClick());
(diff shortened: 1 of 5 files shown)


Artur-:
Is this still relevnt?

mcollovati:
> Is this still relevnt?

I think so because it fixes a real bug.
The current click behavior invokes the Javascript `click` function, basically missing additional information a "normal" click carries in the event object.
A first attempt to change the behavior was reverted because it broke many tests in the flow-components repo.
There was no agreement on how to handle that. This is a second attempt to change the behaviour, but with an opt-out option to preserve the existing behaviour.

However, the discussion stopped at some point

TatuLund:
> I think so because it fixes a real bug.

I agree, we have real customer cases them having problems with this.

PR #2160, opened 2026-03-03
feat: delegate waitForVaadin readiness checks to Flow

Artur-:
Replace the synchronous polling script that checked document.readyState,
devServerIsNotLoaded, and active clients directly with a call to
Flow's whenReady callback, which now owns all readiness logic.

TestBench polls Java-side until whenReady is a function (handling dev
server startup where the page may not have Flow loaded yet), then
makes a single executeAsyncScript call to let Flow report readiness.

Remove the waitForVaadinLoopHook testing mechanism and update
integration tests to use whenReady=false and setTimeout instead.

### DIFF
--- a/vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
+++ b/vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
@@ -8,10 +8,7 @@
  */
 package com.vaadin.tests;
 
-import java.lang.reflect.Field;
-
 import org.junit.jupiter.api.Assertions;
-import org.openqa.selenium.JavascriptExecutor;
 
 import com.vaadin.flow.component.Component;
 import com.vaadin.testUI.PageObjectView;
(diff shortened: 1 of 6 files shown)


vaadin-testbench-shared/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java
```diff
@@ -48,27 +50,16 @@ private static Logger getLogger() {
     private boolean enableWaitForVaadin = true;
     private boolean autoScrollIntoView = true;
     // @formatter:off
-    String WAIT_FOR_VAADIN_SCRIPT =
-            "if (document.readyState != 'complete') {"
```
TatuLund: Do not remove document.readyState check.

mcollovati: I think the plan is to move the readyState check into the new `window.Vaadin.Flow.whenReady` function

https://github.com/vaadin/flow/pull/23739/changes#diff-bc9169eba7c2d4af1e8efaae0fb39a9e9f3a83b4ade049fa0f59571db090ea6b

vaadin-testbench-integration-tests-junit5/src/test/java/com/vaadin/tests/WaitForVaadinIT.java
```diff
@@ -61,41 +58,24 @@ public void waitForVaadin_noConnectors_returnsImmediately() {
...
-        assertExecutionNoLonger(() -> getCommandExecutor().waitForVaadin());
-    }
-
     @BrowserTest
     public void waitForVaadin_devModeNotReady_waits() {
         openTestURL();
-        getCommandExecutor().executeScript(
-                "window.Vaadin = {Flow: {devServerIsNotLoaded: true}};");
+        getCommandExecutor()
+                .executeScript("window.Vaadin.Flow.whenReady = false;");
```
mcollovati: With this setting the test always fails because `waitForVaadin` almost immediately returns, since there's nothing waiting for the timeout expected by `assertExecutionBlocked`

PR #2186, opened 2026-04-09
fix: Failing test should fail recording

caalador:
If a recording test fails on
Failure or Error the recording
should stop on an exception.


tltv (approved):
This change should go to the playwright branch too.

caalador:
Playwright should be rebased on master and at some point merged so it will get the fix also.

PR #2188, opened 2026-04-09
chore: move loadtest-helper content to loadtest-support

caalador:
Move the loadtest-helper functionality
into the vaadin-loadtest-support module
to simplify usage as you now only need
to add a single test dependency.


vaadin-testbench-loadtest/load-tests/demo-web-app/pom.xml
tltv: Shouldn't this be now in `compile` scope (default) to keep the custom runtime error handling work?

caalador: That is true. Good catch. Changed scope

tltv (approved):

PR #2190, opened 2026-04-10
feat: Add custom thresholds

caalador:
Add possibility to set custom
thresholds for metrics.


vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/ThresholdConfig.java
```diff
@@ -45,24 +150,54 @@ public record ThresholdConfig(int httpReqDurationP95, int httpReqDurationP99,
...
             sb.append(
                     "    checks: [{ threshold: 'rate==1', abortOnFail: true, delayAbortEval: '5s' }],\n");
         } else {
             sb.append("    checks: ['rate==1'],\n");
         }
-        sb.append("    http_req_duration: [");
-        boolean first = true;
+
+        // http_req_duration: merge defaults with any custom expressions
+        List<String> durationExpressions = new ArrayList<>();
```
tltv, line 163: ArrayList is not checked for duplicates. How about changing it to HashSet or LinkedHashSet to avoid duplicates in the list.

caalador: It only gets the specific `p(95)<`, `p(99)<` (these from plugin config) and `http_req_duration` from the custom string and the custom is a map and can not contain duplciates for the http_req_duration.

tltv: Maybe there could be a test for `ThresholdConfig.DEFAULT.withCustomThresholds("http_req_waiting:p(95)<2000,http_req_waiting:p(99)<5000")` case to verify that since looking at the code it's not so obvious.

caalador: test added

tltv: I meant `http_req_duration`, my bad. No need for new test for `http_req_waiting`. But a test for overlapping with the defaults like this would be nice:
`ThresholdConfig.DEFAULT.withCustomThresholds("http_req_duration:p(95)<2000,http_req_duration:p(99)<5000")`.

I tried it now myself and looks like `toK6ThresholdsBlock()` would indeed produce duplicate entries.

caalador: Now duration handles custom p95 and p99 as overriding the defaults generating no duplicates to the output.

tltv (approved):

PR #2193, opened 2026-04-10
feat: Load profile definition

caalador:
Add possibility to define
load pattern for k6 test execution


tltv (commented):
This feature could be a bit more configurable. Currently it doesn't let you choose the executor (there are many more as listed [here](https://grafana.com/docs/k6/latest/using-k6/scenarios/executors/)). And also you can only customize stages, but nothing else like `startVUs`, or `gracefulRampDown` and many more. It's nice to have pre-defined profiles, but at the same time the API could allow customizing the executor and custom properties.

caalador:
Extended configurability and updated documentation.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/LoadProfile.java
```diff
@@ -0,0 +1,822 @@
...
+     * @param scenarioContent
+     *            raw k6 JavaScript scenario properties
+     * @return a LoadProfile that inserts the content directly
+     */
+    public static LoadProfile customScenario(String scenarioContent) {
+        return new LoadProfile(null, null, "0s", "0s", List.of(), null, null,
+                null, null, null, null, scenarioContent);
+    }
+
+    private LoadProfile(LoadPattern pattern, K6Executor executor, String rampUp,
```
tltv: This constructor is too big. I would remove it and change necessary class members to non-final, add `public static LoadProfile executor1(x, y, z)`, `public static LoadProfile executor2(x, y)` etc. instead. In similar way as with customScenario, but without any huge constructors. Having one smaller to start with and then filling in individual fields as needed would be cleaner to use and read: 

```
public static LoadProfile executor1(x, y) {
  LoadProfile p = new LoadProfile(a, b, c, d); 
  p.x = x; 
  p.y = y;
  return p;
}
```

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
```diff
@@ -154,6 +157,140 @@ public class K6RunMojo extends AbstractK6Mojo {
...
+     * ramping-vus executor)</li>
+     * <li>{@code stress} - gradual increase with a 150% spike phase</li>
+     * <li>{@code soak} - quick ramp, extended sustain for long-duration
+     * tests</li>
+     * <li>{@code custom} - user-defined stages via {@code k6.stages}</li>
+     * </ul>
+     *
+     * @see LoadProfile.LoadPattern
+     */
+    @Parameter(property = "k6.loadPattern", defaultValue = "ramp")
```
tltv, line 176: Should it be ramp or constant by default? K6ScenarioCombiner has LoadProfile.CONSTANT by default in couple constructors, but here it's ramp by default. Also there's LoadProfile.DEFAULT, but why that's not used anywhere?

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/LoadProfile.java
caalador: Moved to builder and factories.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
caalador: Ramp is the better one I would say and if it brings problems it is easy to switch to constant.
Dropping the only test DEFAULT profile.

tltv (approved):

PR #2194, opened 2026-04-13
fix: csv to handle , and " in input

caalador:
Correctly handle values containing
, or " inside the value.
Previously , inside value was not
correctly handled by the csv reader
even though the recording correctly
"" marked the value.


mshabarov (approved):
May want to fix the copilot's comment, otherwise no other blockers.

mshabarov (approved):

PR #2197, opened 2026-04-17
feat: collect results into json

caalador:
Collect run results into
json file.
Extra collected data that
is not shown in console are,
per-request breakdowns.

### DIFF
--- a/vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
+++ b/vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
@@ -308,6 +308,10 @@ public class K6RunMojo extends AbstractK6Mojo {
         // Initialize (extract utilities)
         initialize();
 
+        // Configure summary stats to match threshold percentiles
+        nodeRunner.setSummaryTrendStats(
+                buildThresholdConfig().toSummaryTrendStats());
+
         // Validate k6 is available
         if (!nodeRunner.isK6Available()) {
             throw new MojoExecutionException(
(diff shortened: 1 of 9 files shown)


vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
```diff
@@ -60,6 +60,7 @@ public class HarToK6Converter {
             import http from 'k6/http'
             import { check, fail } from 'k6'
             import { sleep } from 'k6'
+            import { textSummary } from 'https://jslib.k6.io/k6-summary/0.0.1/index.js'
```
tltv: This file could be stored locally in k6-utils folder to avoid fetching from external source and added with a comment where to get the original source:

`curl -o src/main/resources/k6-utils/k6-summary.js https://jslib.k6.io/k6-summary/0.0.1/index.js`

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
```diff
@@ -308,6 +308,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         // Initialize (extract utilities)
         initialize();
 
+        // Configure summary stats to match threshold percentiles
+        nodeRunner.setSummaryTrendStats(
```
tltv, line 312: Should we have maven property for this? I was looking for it to be able to show a bit less metrics by default in the console output.

tltv (changes requested):

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
caalador: copied local version

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/K6RunMojo.java
caalador: made metrics short for console.
Keeping else as is as we expect data to be available.

tltv (approved):

PR #2201, opened 2026-04-17
feat: Report as html

caalador:
Generate a html report on the
executed loadtest(s).

builds on #2197


vaadin-testbench-loadtest/testbench-converter-plugin/src/main/resources/report/summary-report-generated.html
tltv, line 1: This file should probably say in the beginning that: "This file is generated by Claude" etc. Or file name could end with -aigenerated.html. Just to make it clear that it's deliberately not human-readable.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/NodeRunner.java
```diff
@@ -509,6 +533,163 @@ public void runK6Test(Path testFile, int virtualUsers, String duration,
...
+     * <li>{@code request_timeline} — per-second buckets with request count,
+     * pass/fail counts, and avg/max response time</li>
+     * </ul>
+     *
+     * @param csvFile
+     *            the k6 CSV metrics output file
+     * @param summaryFile
+     *            the summary JSON file to inject the data into
+     */
+    private void injectCsvMetrics(Path csvFile, Path summaryFile) {
```
tltv: This whole method would be better in some K6 CSV ouput+JSON summary specific class. It doesn't make much sense without knowing exactly what's the input CSV format and JSON format for the output.

tltv (commented):
In the output html, in the request table, requests are cut off. I understand it's mainly due to the same value being used as part of a "tag" in the json data. Could it be longer? And cut it in the middle to show the ending (like `/VAADIN/build/a...defgh12345.html`), which is often the interesting part.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/resources/report/summary-report-generated.html
caalador: As far as pure html +javascripr with json handling and svg generation goes I would actually say this is quite readable, but added comment.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/NodeRunner.java
caalador: moved, expanded usage for javadoc

caalador:
Made tag show more characters, but concatenate the chunk hex

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/resources/report/summary-report-generated.html
tltv: Now it ends up in the end result file too which is not necessary. To keep it simple just naming the template file differently would work better, but this is not a blocker.

caalador: renamed to generated. doubt it matters how it is/was generated.

tltv, line 1: Thresholds table has Metric column has wrong value. Its always static text: "TestBench Converter Maven Plugin".

tltv (commented):
Add some basic unit test that verify that summary json and html files are generated when expected. Preferably ResultHandler has a test that verifies the full expected json content based on single request and other data.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/resources/report/summary-report-generated.html
caalador: Maven replaced too much. Fixed.

tltv (approved):

PR #2205, opened 2026-04-21
test: wait for sauce labs

caalador:
Wait for sauce labs to
be ready before starting
integration tests

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -191,6 +191,29 @@ jobs:
           retryTimeout: 300
           proxyLocalhost: allow
 
+      - name: Wait for Sauce Labs tunnel to be ready
+        env:
+          SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+          SAUCE_TUNNEL_ID: ${{ github.run_id }}-${{ github.run_number }}
+        run: |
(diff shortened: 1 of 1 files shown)


tltv (approved):
nit: swap the grep/|| true pair for jq + proper curl error handling, and drop a one-line comment on why the sauce-connect-action's own readiness isn't trusted.

Details:

- Relationship to `retryTimeout: 300` on sauce-connect-action is unexplained. The preceding `saucelabs/sauce-connect-action@v3.0.0` already has `retryTimeout: 300`. If that step returns before the tunnel is actually usable, this polling step is papering over an upstream issue. Worth a one-line comment stating *why* the action's own readiness signal isn't sufficient (flaky? returns early? race with API propagation?) — otherwise a future reader will either remove this as redundant or duplicate it elsewhere.

- Fragile JSON parsing. `grep -o "\"tunnel_identifier\":\"${SAUCE_TUNNEL_ID}\""` depends on literal-character adjacency in the response (no whitespace, no reordering, no escaping). `jq` is pre-installed on GitHub runners — safer:


```
curl -sSf -u "$SAUCE_USERNAME:$SAUCE_ACCESS_KEY" "$URL" \
      | jq -e --arg id "$SAUCE_TUNNEL_ID" 'any(.tunnel_identifier == $id)'
```


- `|| true` hides real failures. If curl fails (5xx, 401, DNS blip), `STATUS` is empty and the loop logs "not ready yet" indistinguishably from "tunnel genuinely not listed." Consider `curl -sSf` and checking exit code separately so transient API errors don't eat the entire 150s budget silently.

PR #2214, opened 2026-04-23
feat: add new playwrightHelper with context

### LINKED ISSUE #2209: Add method to PlaywrightHelper
if the user defines their own Browser.NewContextOptions then the current `PlaywrightHelper.createBrowserContext(Browser)` doesn't use that in any way with or without the `k6.harOutputPath`

Add new method `PlaywrightHelper.createBrowserContext(Browser, Browser.NewContextOptions)` that uses given contextOptions to generate the newContext.

caalador:
Add a new createBrowserContext for
PlaywrightHelper which takes in
a user defined NewContextOptions.

Add tests for helper class.

Fixes #2209

### DIFF
--- a/vaadin-testbench-loadtest/testbench-loadtest-support/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
+++ b/vaadin-testbench-loadtest/testbench-loadtest-support/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
@@ -1,0 +1,1 @@
+mock-maker-subclass
--- a/vaadin-testbench-loadtest/testbench-loadtest-support/pom.xml
+++ b/vaadin-testbench-loadtest/testbench-loadtest-support/pom.xml
@@ -71,6 +71,26 @@
             <version>${junit.version}</version>
             <scope>provided</scope>
         </dependency>
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>5.15.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>5.15.2</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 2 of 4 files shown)


vaadin-testbench-loadtest/testbench-loadtest-support/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-subclass
```
tltv, line 1: Why is this needed in this module?

caalador: Without it we get
```
[ERROR]   PlaywrightHelperTest.createBrowserContext_withOptions_appliesHarRecordingWhenPropertySet:62 Mockito 
Mockito cannot mock this class: interface com.microsoft.playwright.Browser.

If you're not sure why you're getting this error, please open an issue on GitHub.


Java               : 25
JVM vendor name    : Amazon.com Inc.
JVM vendor version : 25.0.2+10-LTS
JVM name           : OpenJDK 64-Bit Server VM
JVM version        : 25.0.2+10-LTS
JVM info           : mixed mode, sharing
OS name            : Windows 11
OS version         : 10.0


You are seeing this disclaimer because Mockito is configured to create inlined mocks.
You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.

Underlying exception : org.mockito.exceptions.base.MockitoException: Could not modify all classes [interface java.lang.AutoCloseable, interface com.microsoft.playwright.Browser]
```
So this is needed for JDK 25 to mock AutoCloseable interfaces

tltv: Is this error coming with corretto-25.0.2? I don't get it at all even without this file with corretto-25.0.3, or temurin 25 or openjdk 25. Just wondering if you hit a jdk bug actually. Could you try with corretto-25.0.3? And if it works, need to check version in CI before removing the file.

tltv: Oh nevermind, I got the MockMaker file locally despite never pulling your changes...

caalador: But updated to `JVM version        : 25.0.3+9-LTS` and fails with the same issue on jdk 25, but if you take for instance JDK24 it passes without the extension

tltv: CI might have older than 25 then. I got this same file generated for my PR too but missed it totally since tests are green locally and in CI.

tltv (approved):

PR #2216, opened 2026-04-23
fix: resolve node IDs at runtime in k6 scripts

tltv:
Recorded server-assigned node IDs drift whenever the UI is reordered or Flow bumps its allocation order. The converter now maps each node to a stable key (id attribute, else tag#ordinal) by walking each UIDL response's `changes` array, and rewrites `"node":N` RPC references as `${resolveNode(nodeMap, 'key')}`. A companion `updateNodeMap` export in `vaadin-k6-helpers.js` keeps the map in sync inside every VU.

Summary
=======

Implemented runtime resolution of Vaadin Flow node IDs in generated k6 scripts.

HarToK6Converter.java
---------------------
- `registerChangesFromResponse` / `registerChange` walk UIDL / init response changes arrays via Jackson, deriving stable keys in priority order id -> tag#ordinal.
- `rewriteResolvableNodeIds` replaces positive "node":N in the escaped body with `${resolveNode(nodeMap, 'key')};` unresolved IDs stay as literals and get a `/* unresolved node N */` diagnostic line above the `http.post(` call.
- Init-block now declares `let nodeMap = {}` + calls `updateNodeMap;` `generateUidlResponseExtractionCode` refreshes nodeMap after each UIDL response; re-extraction after subsequent init (post-login) resets the map.
- Extra import line added to `K6_SCRIPT_IMPORTS`

vaadin-k6-helpers.js
--------------------
- New `updateNodeMap(nodeMap, responseBody)` and `resolveNode(nodeMap, key)` exports mirroring the Java-side key derivation.
- resolveNode throws Error("unresolved node: " + key) on miss.

Tests
-----
- `HarToK6ConverterNodeResolutionTest` with 6 cases: id key, tag+ordinal, duplicate tags, negative IDs, unresolved, opt-out byte-equivalence.


vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/AbstractK6Mojo.java
```diff
@@ -100,6 +100,18 @@ public abstract class AbstractK6Mojo extends AbstractMojo {
...
+     * When {@code true} (the default), generated k6 scripts resolve Vaadin node
+     * IDs at runtime via a stable-key → node-ID map built from each UIDL
+     * response. Setting it to {@code false} emits the legacy behaviour where
+     * {@code "node":N} literals from the recording are baked into the script
+     * and thus break whenever the recorded UI is reordered, re-rendered
+     * conditionally, or a Flow framework bump shifts the ID allocation order.
+     * Keep it enabled unless you are A/B-diagnosing a regression.
+     */
+    @Parameter(property = "loadtest.resolveNodeIds", defaultValue = "true")
+    protected boolean resolveNodeIds;
```
caalador: I would not add the possibility to disable the nodeId mapping as applications break anyway without any regressions.
Also we now need to throw the boolean around alot.

Probably we should remove all the overloaded methods and constructors and use a RecorderOptions that gets passed around instead.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
```diff
@@ -64,6 +80,12 @@ public class HarToK6Converter {
 
             """;
 
+    /**
+     * Extra imports pulled from {@code vaadin-k6-helpers.js} when runtime
+     * node-ID resolution is enabled.
+     */
+    private static final String NODE_RESOLVER_IMPORT = "import { updateNodeMap, resolveNode } from '../utils/vaadin-k6-helpers.js'\n";
```
caalador: These could be just added to the HELPER_IMPORT in K6TestRefactorer as having them there always doesn't change things.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
```diff
@@ -117,14 +153,36 @@ export function handleSummary(data) {
...
 
     /**
-     * Creates a new HAR to k6 converter with a configured Jackson ObjectMapper.
+     * When {@code true}, emits runtime node-ID resolution scaffolding and
+     * rewrites positive {@code "node":N} UIDL payload references as
+     * {@code ${resolveNode(nodeMap, 'stable-key')}}. When {@code false}, the
+     * converter emits the legacy literal-number behaviour — useful as an escape
+     * hatch for unanticipated UIDL shapes.
+     */
+    private final boolean resolveNodeIds;
```
caalador: Would say that this shouldn't be an option.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
```diff
@@ -206,6 +266,9 @@ public void convert(Path harFile, Path outputFile,
 
         StringBuilder script = new StringBuilder();
         script.append(K6_SCRIPT_IMPORTS);
+        if (resolveNodeIds && session != null) {
```
caalador: The imports could just always be added with the others as it will not make much of a difference.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/NodeRunner.java
```diff
@@ -216,9 +216,35 @@ public void harToK6(Path harFile, Path outputFile,
             ThresholdConfig thresholdConfig,
             ResponseCheckConfig responseCheckConfig)
```
caalador: Lets change these to an RecorderOptions object instead of adding more fields.

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/AbstractK6Mojo.java
tltv: done

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
tltv: removed

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
tltv: moved to K6_SCRIPT_IMPORTS

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/HarToK6Converter.java
tltv: done

vaadin-testbench-loadtest/testbench-converter-plugin/src/main/java/com/vaadin/testbench/loadtest/util/NodeRunner.java
tltv: done

caalador (approved):

PR #2225, opened 2026-05-04
chore: upgrade spring boot 4.1.0-RC1

ZheSun88:


pom.xml
```diff
@@ -37,7 +37,7 @@
         <maven.compiler.release>21</maven.compiler.release>
         <maven.compiler.source>${maven.compiler.release}</maven.compiler.source>
         <maven.compiler.target>${maven.compiler.release}</maven.compiler.target>
-        <spring.boot.version>4.0.4</spring.boot.version>
+        <spring.boot.version>4.0.6</spring.boot.version>
```
mshabarov: Why not to 4.1 RC ?

mshabarov: (only for 10.2)

ZheSun88: oops.. true that..

mshabarov (approved):

PR #2228, opened 2026-05-06
feat: Add warmup run

caalador:
Add configuration 'warmup'
that when enabled does a single
iteration with one user
of the test script against
the server before starting the
actual load test run.


tltv (changes requested):
Does this work with default function like: `export default function () {`?

Add simple test e.g. in existing `run-constat-load` IT. As long as default function is not a problem, it should be simple as enabling warmup in `pom.xml` and adding single assert in `verify.bsh` like `IntegrationTestHelper.assertLogContains(basePath, "Warmup completed");`

tltv (approved):

PR #2229, opened 2026-05-06
fix: rebuild form-encoded POST bodies from postData.params

### LINKED ISSUE #2227: HarToK6Converter drops form-encoded POST bodies, breaks Spring Security login replays
## Summary

When a recorded TestBench scenario contains an `application/x-www-form-urlencoded` POST (e.g. Spring Security's `POST /login`), the generated k6 script ships an **empty body** for that request. The recording itself is fine — BrowserMob proxy parses the form fields into `postData.params` — but `HarToK6Converter` only reads `postData.text`, which BrowserMob leaves as `null` for parsed form bodies. As a result, the load test never actually authenticates: Spring Security 302-redirects every iteration to `/login?error`, every VU stays on the login page, and any subsequent UIDL request that references a post-login DOM node (e.g. `vaadin-app-layout#1`) blows up in `resolveNode`.

## Environment

- `testbench-converter-plugin` 10.2.0-alpha4
- `testbench-loadtest-support` 10.2.0-alpha4
- Vaadin 25.2.0-alpha5, Spring Boot 4.0.6, Spring Security 6 (CSRF on by default)
- macOS / Linux, Java 21, k6 (latest)

## Repro

1. Take any Vaadin Flow + Spring Security app that has a `LoginViewElement`-style E2E test driving the form login.
2. Use the testbench-converter-plugin's `record` and `run` goals against that test.
3. Observe `target/k6/tests/<scenario>/report/<scenario>-errors.log` — it will contain one `unresolved node` error per iteration. Inspecting the generated `<scenario>.js` shows:

   ```js
   // Request N: POST http://localhost:8081/login
   response = http.post(
     `${BASE_URL}/login`,
     '',                    // <-- empty body
     { headers: { 'Content-Type': 'application/x-www-form-urlencoded', ... } }
   )
   ```

4. The corresponding HAR entry has the credentials in `postData.params` but `postData.text: null`:

   ```json
   "postData": {
     "mimeType": "application/x-www-form-urlencoded",
     "text": null,
     "params": [
       { "name": "_csrf", "value": "..." },
       { "name": "username", "value": "user" },
       { "name": "password", "value": "password" }
     ]
   }
   ```

## Expected

`HarToK6Converter` reconstructs the request body from `postData.params` (URL-encoded `name=value&name=value...`) when `postData.text` is null/empty and `mimeType` is `application/x-www-form-urlencoded`.

## Actual

`com.vaadin.testbench.loadtest.util.HarToK6Converter$HarPostData` is a record of `(mimeType, text)` only — `params` is never read. The generator emits `''` for the body, the load test silently degrades to a redirect loop, and `http_req_failed` stays at 0% because 302s count as success.

## Suggested fix

In `HarToK6Converter`:

1. Extend `HarPostData` (or its deserialization) to carry `params` as well as `text`.
2. When emitting the request body, fall back to `URLEncoder.encode`-ing `params` (in declaration order) if `text` is null/empty and the content type is form-encoded.
3. For Spring Security flows specifically, consider templating common session-bound fields like `_csrf` so each VU uses its own session's token instead of replaying the recorded one. Spring rotates the CSRF token after each auth attempt, so the replay needs to re-extract from the latest login-page HTML rather than reuse a single recorded value.

## Secondary observation

The `record` goal also passes `-Djunit.jupiter.extensions.autodetection.enabled=true` to failsafe and expects `K6RecordingExtension` to register via ServiceLoader — but the current error message ("Proxy recorded 0 requests... Verify K6RecordingExtension is on the classpath") points at the wrong cause when the real problem is that the test never called `LoadTestItHelper.openWithProxy(...)`. Could the message also suggest "make sure your `@BeforeEach` opens the URL via `LoadTestItHelper.openWithProxy(getDriver(), ...)`"?

## Workaround

Until the converter is fixed, post-process the generated k6 script between `record` and `run`. We wired a Python script into Maven via `exec-maven-plugin` in `post-integration-test`, with `run-load-tests` and `stop-server` moved to the `verify` phase so the patch lands between recording and replay. The script:

1. Walks the recorded HAR, collects every POST whose `postData.text` is null but whose `postData.mimeType` is form-encoded with non-empty `params`.
2. Rewrites the matching `http.post(\`${BASE_URL}<path>\`, '', ...)` in the generated `.js`, URL-encoding the params back into the body.
3. For the `_csrf` field specifically, emits `${hillaCsrfToken}` instead of the recorded value, widens the script's `const hillaCsrfToken` to `let`, and injects `hillaCsrfToken = getHillaCsrfToken(response.body) || hillaCsrfToken;` after every response check so the token tracks Spring Security's per-attempt rotation.

Full workaround lives in this repo: `loadtest/scripts/patch-form-bodies.py` plus the `exec-maven-plugin` execution in `loadtest/pom.xml`.

After applying the workaround on a 1000-VU / 30s run against a Vaadin 25 + Spring Security app:

|  | Before workaround | After workaround |
|---|---|---|
| `unresolved node` errors | 2,289 (1 per iter) | 0 |
| `iteration_duration` avg | ~10s (aborted at login) | ~26s (full flow incl. grid render) |
| `http_req_duration` p95 | 0s | 204 ms |
| `valid UIDL response` checks | only pre-login UIDLs | includes post-login `vaadin-app-layout` interactions |

**Affected version:** 
`testbench-converter-plugin` 10.2.0-alpha4 (and the bundled `testbench-loadtest-support`)

tltv:
BrowserMob parses form fields into postData.params and leaves postData.text null, so the converter shipped Spring Security login posts with an empty body and every VU stayed on /login?error.

HarPostData now carries params; when text is null/empty and the mime type is application/x-www-form-urlencoded the body is URL-encoded from the recorded params in declaration order. The recorded _csrf value is replaced with ${hillaCsrfToken} so each VU sends its own session's token.

Fixes: #2227


tltv:
Verified to work with https://github.com/vaadin/bookstore-example/tree/v25-loadtest

caalador (approved):

PR #2233, opened 2026-05-07
fix: Improve viewport resizing behavior (backport of #2170)

thevaadinman:
Original description:

The fix replaces the single-shot adjustment with a converging retry loop. The browser chrome (title bar, borders) is constant for a session. Iteration 1 discovers the offset, iteration 2 (or rarely 3) nails the target. 5 attempts is generous safety margin for unusual environments (remote grids, DPI scaling).

Removes hardcoded Mac FF offsets that are irrelevant for modern browsers and cause unnecessary overshoot on other platforms (extrah=106, extraw=0) — starts with the desired dimensions directly

Loops up to 5 times, each iteration measuring the actual viewport via JS, computing the diff, and adjusting the window size using the current window size (not accumulated offsets)

Returns early as soon as the viewport matches the target

Re-throws UnsupportedOperationException without wrapping, while wrapping other exceptions

Logs each adjustment at debug level for troubleshooting

Fixes: [#2167]

### DIFF
--- a/vaadin-testbench-core/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java
+++ b/vaadin-testbench-core/src/main/java/com/vaadin/testbench/commands/TestBenchCommandExecutor.java
@@ -292,33 +292,48 @@ public class TestBenchCommandExecutor
     @Override
     public void resizeViewPortTo(final int desiredWidth,
             final int desiredHeight) throws UnsupportedOperationException {
+        final int MAX_RESIZE_ATTEMPTS = 5;
         try {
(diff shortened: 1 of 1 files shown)


thevaadinman:
Test failures are unrelated to this patch.. the test failures are due to the versions of various testing librarires that TestBench 5.4 uses being incompatible with Java 21+, which GitHub Actions uses. This was not an issue with the old CI setup that ran on TeamCity since we had it configured to build on a Java 8 JDK.
This PR is not the correct place to fix those test failures.

Ansku (approved):

PR #2242, opened 2026-05-08
fix: anchor gridKeys regex to grid item rows in generated k6

tltv:
The previous regex `/"key":"[^"]+"/g` matched both grid item rows `({"key":"1","col0":...})` and change-record keys `({"node":N,"type":"put","key":"highlight",...})`. On responses with no grid items it overwrote `gridKeys` with property names like "highlight" or "tag", causing select() RPCs to send garbage keys and downstream resolveNode() throws (e.g. "unresolved node: abc"). Anchoring on `{` matches only keys that are the first property of an object, so grid items are picked up and change-record keys are ignored; when no items are present, gridKeys keeps its previous valid contents.

Replace the naive walker in `K6ScenarioCombiner`. The combiner is now agnostic to braces in human-written comments and string literals.

### DIFF
--- a/vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6ScenarioCombinerTest.java
+++ b/vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6ScenarioCombinerTest.java
@@ -206,6 +206,46 @@ class K6ScenarioCombinerTest {
     }
 
     @Test
+    void combine_scenarioWithUnbalancedBracesInRegexAndCommentsClosesCleanly()
+            throws IOException {
+        // The brace walker must skip braces inside regex literals, comments,
+        // and strings. Each line of the body below contains a shape that
+        // broke the previous naive counter (regex `\{`, isolated `{` in a
+        // comment, isolated `{` and `}` in string literals); without
+        // lexer-aware skipping, the walker overshoots and slurps the
+        // trailing handleSummary export.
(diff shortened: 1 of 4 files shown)


tltv:
Verified to work with https://github.com/vaadin/bookstore-example/tree/v25-loadtest.

vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6ScenarioCombinerTest.java
```diff
@@ -205,6 +205,46 @@ void equalWeights_convertsKebabCaseToCamelCaseName() throws IOException {
...
+        int handleSummaryCount = 0;
+        int idx = 0;
+        while ((idx = content.indexOf("function handleSummary", idx)) != -1) {
+            handleSummaryCount++;
+            idx += "function handleSummary".length();
+        }
+        assertEquals(1, handleSummaryCount,
+                "Combined script must export exactly one handleSummary. Got:\n"
+                        + content);
+        assertTrue(content.contains("export function alphaScenario()"),
```
caalador: This doesn't really check that the function is correctly wrapped, only that the function export has been added with the file name. should probably check the body given to writeScenario is there, or all the lines are found as full lines.

tltv: Added scenario wrapper body line asserts.

caalador (approved):

PR #2258, opened 2026-05-25
fix: ignore parens inside think-time sleeps

### LINKED ISSUE #2253: Loadtest thinkTime broke combined-scenarios.js
Keeping think time enabled will break combined-scenarios.js. Thinking times are inserted in wrong positions.

Reproducible with https://github.com/vaadin/bookstore-example/tree/v25-loadtest when `thinkTimeEnabled` is true. Testbench versions 10.2.0-alpha4 and alpha5.

tltv:
`K6TestRefactorer` counted `(` and `)` with no string-awareness, so a literal `)` inside a header value (e.g. Chrome `sec-ch-ua` `"Not/A)Brand";v="99")` made parenCount drop to zero mid-headers. The think-time sleep() was then injected inside the headers object, breaking JS parsing in `combined-scenarios.js` when `combineScenarios=true`. Same fragility applied to brace-counting in JSESSION/Vaadin extraction insertion.

Replace countOccurrences with countUnquotedOccurrences that tracks `'..'`, `".."` and ``` `.` ``` string state and honors backslash escapes.

Fixes: #2253

### DIFF
--- a/vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6TestRefactorerTest.java
+++ b/vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6TestRefactorerTest.java
@@ -145,7 +145,7 @@ class K6TestRefactorerTest {
     }
 
     @Test
-    void refactorContent_thinkTimesEnabled_headerValueContainsCloseParen_sleepStaysAtStatementScope() {
+    void refactorContent_thinkTimesEnabled_headerValueContainsCloseParen_sleepIsNotInjectedInsideHeaders() {
         K6TestRefactorer refactorer = new K6TestRefactorer(
                 ThinkTimeConfig.DEFAULT);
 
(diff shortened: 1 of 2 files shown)


tltv:
Verified to fix the issue in https://github.com/vaadin/bookstore-example/tree/v25-loadtest.

vaadin-testbench-loadtest/testbench-converter-plugin/src/test/java/com/vaadin/testbench/loadtest/util/K6TestRefactorerTest.java
```diff
@@ -144,6 +144,102 @@ export default function() {
...
+        // is string- and line-comment-aware so it isn't fooled by `)` inside
+        // header values or `(` inside `// Request N:` comments.
+        int functionBodyOpen = refactored
+                .indexOf("export default function() {");
+        assertTrue(functionBodyOpen >= 0,
+                "Function body not found in refactored output:\n" + refactored);
+        int scanFrom = functionBodyOpen
+                + "export default function() {".length();
+
+        int parenBalance = 0;
```
caalador: The verification is a bit complicated and would sound like a better approach would be to check that the expected sleep() lines exist and that they do not break the `Not/A)Brand`

caalador (approved):

PR #2261, opened 2026-05-29
refactor: deprecate `$` and `$view` in favor of `find` and `findInView`

mcollovati:
The `$` and `$view` query methods in the UI unit test base class have been deprecated, aligning the API with the browserless-test replacement.

Introduce `find` and `findInView` as their replacements and update the tests to use the new methods.

### DIFF
--- a/vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
+++ b/vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
@@ -452,7 +452,8 @@ public abstract class BaseUIUnitTest {
      *            the type of the component(s) to search for
      * @return a query object for finding components
      */
-    public <T extends Component> ComponentQuery<T> $(Class<T> componentType) {
+    public <T extends Component> ComponentQuery<T> find(
+            Class<T> componentType) {
         verifyAndGetUI();
         return internalQuery(componentType);
     }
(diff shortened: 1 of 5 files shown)


vaadin-testbench-unit-shared/src/main/java/com/vaadin/testbench/unit/BaseUIUnitTest.java
```diff
@@ -484,14 +485,66 @@ private static <Y extends Component> Class<? extends ComponentTester> getTester(
...
+     *            the type of the component(s) to search for
+     * @return a query object for finding components
+     * @deprecated since 25.2, for removal; use {@link #findInView(Class)}
+     *             instead.
+     */
+    @Deprecated(since = "25.2", forRemoval = true)
+    public <T extends Component> ComponentQuery<T> $view(
+            Class<T> componentType) {
+        return findInView(componentType);
+    }
```
mshabarov, line 546: WHy do we add these new methods and deprecate? Wasn't it so that these $ methods exist?

mcollovati: They are not new methods. It seems so only because the "old" methods has been renamed, so we need to "re-add" them

mshabarov: Right, thanks.

mshabarov (approved):

PR #2312, opened 2026-09-10
test: fix test assertion

mcollovati:
Theme attribute values are now in a fixed order that does not match the test expectation. This change fixes the assertion.


ZheSun88 (approved):

ZheSun88:
nice  i was just waiting for this..

